Issue #2, opened 2019-04-06
RFC: Introduce means to "destroy"/"remove" metrics objects


ktoso:
Importing from https://github.com/tomerd/swift-server-metrics-api-proposal/issues/6

Rewritten the rationale in a more formal form, please discuss.

--- 
# Define a way to "destroy" metrics

- Previous ticket: https://github.com/tomerd/swift-server-metrics-api-proposal/issues/6
- Status: implemented https://github.com/apple/swift-metrics/pull/3 open to adjustments
- For end-users: This is an additive change, it enables end-users to do things which they were not able to do before; it does not _force_ them to do so.
- For metrics library authors: This sets a strong implementation style recommendation, that resources shall be allowed to be destroyed at the users request. We ask that implementations important for the Swift Server ecosystem adopt this semantic, yet also allow that there may be implementations which do not have a need for implementing this call (e.g. stateless ones)

This proposal aims to explain the need for "destroying" metrics via the common Metrics API. The proposal explores a wide variety of possible metrics library implementations, and attempts to strike a balance that is _safe_, _performant_ and should not block any implementation from achieving it's specific goals.

In order to bring more focus onto the discussion -- as it has been laying around, but perhaps its importance underestimated so far, the proposal will follow swift-evolution inspired format for the writeup.

## Proposal

There _must_ be a well-defined way to "destroy", metrics in long lived processes if we know a metric will never again be touched.

This proposal forms a case for why this API is required and best to be included right away, rather than as an after thought, and that it is best to perform the "destroy" operations explicitly rather than implicitly.

Having that said, the primary goal here is that there _must_ be a common agreed way to solve this for all implementations of this API, otherwise the model breaks down for any external library exporting metrics -- so if someone had better ideas, we are open to discussing them, within reasonable timeline.

### Terms

Before diving into the case analysis, let's define some terms to be on the same page when we talk about things.

- *metrics API* -- for the sake of this discussion, this entire library apple/swift-metrics
- *metrics backend*, *metrics library* -- a library which implements this API,
- *external library* / "3rd party library" -- for sake of this discussion I mean this only as "some library which depends on metrics API, and wants to offer metrics. It can not and does not know about the *metrics library*." This is _not_ about the 1st/3rd party thing in the sense of "an Apple library or not."
- *short lived metric* -- the "short" is relative here, but the point is "lives shorter than the entire lifetime of the application

### Analysis of cases where delete/destroy does _not_ matter

[...]

PR #3, opened 2019-04-06
 Add MetricsFactory.remove(Metric) to allow metric lifetime management #2

ktoso:
Implements #2 
Based on top of https://github.com/apple/swift-metrics/pull/1 which should land without issues I think, so saving this PR some rebasing work 

Rebased and adjusted from https://github.com/tomerd/swift-server-metrics-api-proposal/pull/11

Feedback welcome, see for more details in the previous PR as well as ticket https://github.com/tomerd/swift-server-metrics-api-proposal/issues/6 as well.

Note that we now don't have a [caching one](https://github.com/apple/swift-metrics/commit/3ee8da0728bfed9292eeacc531ce7b1143890d43) since it was only in the examples, and that's the type of implementations where the release matters the most. So the test is a bit weird, but since the test metrics store internally anyway the test does cover that the right things happen, even if it does not return the "cached one"

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -29,14 +29,18 @@ public protocol CounterHandler: AnyObject {
     /// - parameters:
     ///     - by: Amount to increment by.
     func increment(by: Int64)
+
     /// Reset the counter back to zero.
     func reset()
 }
 
+/// All metric types conform to this protocol, allowing for APIs accepting "any metric."
+public protocol Metric {}
+
 /// A counter is a cumulative metric that represents a single monotonically increasing counter whose value can only increase or be reset to zero.
 /// For example, you can use a counter to represent the number of requests served, tasks completed, or errors.
 /// This is the user facing Counter API. Its behavior depends on the `CounterHandler` implementation
-public class Counter {
+public class Counter: Metric {
     @usableFromInline
     var handler: CounterHandler
     public let label: String
@@ -117,7 +121,7 @@ public protocol RecorderHandler: AnyObject {
 
 /// A recorder collects observations within a time window (usually things like response sizes) and *can* provide aggregated information about the data sample, for example count, sum, min, max and various quantiles.
 /// This is the user facing Recorder API. Its behavior depends on the `RecorderHandler` implementation
-public class Recorder {
+public class Recorder: Metric {
     @usableFromInline
     var handler: RecorderHandler
     public let label: String
@@ -207,7 +211,7 @@ public protocol TimerHandler: AnyObject {
 /// A timer collects observations within a time window (usually things like request durations) and provides aggregated information about the data sample.
 /// For example min, max and various quantiles. It is similar to a `Recorder` but specialized for values that represent durations.
 /// This is the user facing Timer API. Its behavior depends on the `TimerHandler` implementation
-public class Timer {
+public class Timer: Metric {
     @usableFromInline
     var handler: TimerHandler
     public let label: String
@@ -319,6 +323,7 @@ public protocol MetricsFactory {
     ///     - label: The label for the CounterHandler.
     ///     - dimensions: The dimensions for the CounterHandler.
     func makeCounter(label: String, dimensions: [(String, String)]) -> CounterHandler
+
     /// Create a backing `RecorderHandler`.
     ///
     /// - parameters:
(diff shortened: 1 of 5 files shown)


Sources/CoreMetrics/Metrics.swift
```diff
@@ -319,19 +323,46 @@ public protocol MetricsFactory {
...
+    ///
+    /// It is expected that some metrics libraries, may choose to omit implementing this functionality.
+    /// One such example may be a library which directly emits recorded values to some underlying shared storage engine,
+    /// which means that the `MetricHandler` objects themselves are light-weight by nature, and thus no lifecycle
+    /// management and releasing of such metrics handlers is necessary.
+    func remove<M: Metric>(metric: M)
+    // alternatively, delete the Metric type and provide overloads:
+    // func remove<C: Counter>(_ counter: C)
+    // func remove<R: Recorder>(_ recorder: R)
+    // func remove<T: Timer>(_ timer: T)
```
tomerd: i slightly prefer to be symmetric with the `makeXXX(label:)` APIs and offer `destroyXXX(label)` instead of the generic version. this could also allow us to get rid of the `Metric` protocol to keep things simple(r). wdyt?

Sources/CoreMetrics/Metrics.swift
```diff
@@ -448,6 +486,10 @@ public final class NOOPMetricsHandler: MetricsFactory, CounterHandler, RecorderH
         return self
     }
 
+    public func remove<M: Metric>(metric: M) {
+        // no-op
```
MrLotU: Why not leave this on just one line like the other no-op's
```swift
public func remove<M: Metric>(metric: M) { }
```

Mordil:
```swift
public func remove<M: Metric>(metric: M) { /* no-op */ }
```
To even make it explicit that this is intentional

Sources/CoreMetrics/Metrics.swift
```diff
@@ -319,19 +323,49 @@ public protocol MetricsFactory {
...
+    /// However, the presence of this API allows middle-ware libraries wanting to emit metrics for resources with
+    /// well-defined life-cycles to behave pro-actively, and signal when a given metric is known to not be used anymore,
+    /// which can make an positive impact with regards to resource utilization in case of metrics libraries which keep
+    /// references to "heavy" resources.
+    ///
+    /// It is expected that some metrics libraries, may choose to omit implementing this functionality.
+    /// One such example may be a library which directly emits recorded values to some underlying shared storage engine,
+    /// which means that the `MetricHandler` objects themselves are light-weight by nature, and thus no lifecycle
+    /// management and releasing of such metrics handlers is necessary.
+    func release<M: Metric>(metric: M)
```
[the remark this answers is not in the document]
ktoso: The summary of all this is now in the issue #2 so closing here to avoid duplicate discussion threads

Sources/CoreMetrics/Metrics.swift
ktoso: cool, will do 👍

ktoso: (not sure if a formatter like swift format won't "really want to make it 2 lines" but we'll see :))

Sources/CoreMetrics/Metrics.swift
tomerd: @ktoso after sleeping on this some i now even more in favor of the above suggestion for the following reasons:
1. api symmetry 
2. no need in marker protocol 
3. less coupling and confusion between the handler type that the backend libraries deal with (eg CounterHandler) and the metrics object which is what the apps and libraries deal with (eg  Counter). iow, the backend libraries should only deal with handler types, and as such it means a release function that takes a metric object will need to poke a hole  in the abstraction which is not very nice 
4. it opens the door to releasing by an expression which could be useful for middleware type libraries that have a bunch of metrics per connection or request and a naming convention around that

weissi: I'm also in favour of anything that doesn't use the weird marker protocol. `makeXXX` and `destroyXXX` sound good to me!

weissi: or obviously `myMetric.destroy()`, possibly I like this even more...

ktoso: Hmm, some good points, though I'm worried about the `destroyCounter(label: "bla")` losing type-safety and forces more of the "get it right" dance onto users / libraries where as a `destroyCounter(counter)` helps to get it right by compilation; 

Example:

- imagine if you refactor a metrics from counter to gauge `let myThingy = Gauge(...) // was Counter` -- there you had to since you want to get the diff type
- you could forget to change the `destroyCounter(myThingy.label)` call, even if you changed the make site

With accepting `destroyCounter(myThingy)` you would get a compile error and would notice that you'd almost messed up, but the compiler helped you spot that.

ktoso: Hah, I now also realized that the same was true for the `release(Metric)`, so perhaps if we like the 3 functions for the right types that would be a win-win situation?

Sources/CoreMetrics/Metrics.swift
ktoso: Done

Sources/CoreMetrics/Metrics.swift
tomerd: @ktoso you make a great point regarding type safety. however point 3 remains an challenge for me. imo, the application/middleware that will be doing the "releasing" (call site) should only ever know about the metric objects (eg `Counter`, `Timer`) and the backend library should only ever know about the handlers (eg `CounterHandler`, `TimerHandler`) and the proposal atm breaks this separation of concerns. the one common things between them is the label, which is why i arrived at my alternative proposal of releasing by label

@weissi proposal can work too, because the metric object (eg `Counter`, `Timer`) has direct access to the underlying handler (eg `CounterHandler`, `TimerHandler`) by definition so the separation of concerns remains intact

tomerd: /cc @hartbit which had good feedback on logging API, maybe can help here too

ktoso: Hmm, I see what you mean -- that if I don't release I (user) only touch Counter, but if I want to release I suddenly need to reach into all the way to Factory. 

Let me see how it could look like. Main concern was the requirement of additional storage for the reference to the factory, but perhaps -- and also since we enforce that "bootstrap only once" deletes could reach through the shared field...? A bit iffy... could cause weird issues of "release in wrong logging library instance" happening 🤔 

I'll give it a shot as separate commit and we can look at it here

hartbit: Is there a reason we don't benefit from the fact that all `...Handler` protocols are reference types (`AnyObject`) to ask backend libraries to do their cleanup in `deinit`?

ktoso: Thanks for chiming in @hartbit :)
Sadly the `deinit` style has a lot of insane corner cases, and is somewhat abusing the mechanism (the lifecycles of "the metric" do not necessarily match 1:1 "the object"'s lifecycle), I explored it in the writeup over here: https://github.com/apple/swift-metrics/issues/2

I think we are getting to a final design here slowly... we'll be able to:

- keep full typesafety of the releases -- no need to cast
- no crossing boundaries between handler+factory (impl) and counter+system (user api)
- destroy() _can_ live on the objects themselves, leading to simple use and discovery of the feature
- only place to care about "implementing destroying" remains factory, which semantically makes sense

hartbit: Thanks for the explanation! It's true that ARC does come with design challenges, but I'm also worried about a `destroy` function:

* Because destroy will sometimes be a no-op, sometimes not (depending on the implementation), users might misuse them but think everything is fine (because they are using a backend that is a no-op) but start having issues as soon as they switch backends.
* A destroy function doesn't protect users from continuing to use the metrics objects after being destroyed.

While using ARC might mean abusing it slightly, and there cause slightly more headaches in the implementation aspect of things, I think it comes with greater simplicity for the user.

tomerd: /cc @weissi who was part of the previous ARC discussion

PR #12, opened 2019-04-11
fix linux tests formatting

tomerd:
motivation: use consistent formatting

changes: update generate_linux_tests to match the formatting rules we use: `swiftformat --self insert --patternlet inline --stripunusedargs unnamed-only --comments ignore`


tomerd:
fixed via #13

PR #17, opened 2019-04-16
Introduce metric.destroy() to enable lifecycle management #2

### LINKED ISSUE #2: RFC: Introduce means to "destroy"/"remove" metrics objects

ktoso:
Replaces https://github.com/apple/swift-metrics/pull/3
Resolves #2 

// And making it as new PR since I think due to some weird branch magic (since the repo I PR from does not recognize that it is a "fork" in github and I keep getting 500 pages "sometimes" when viewing the PR 3... Reported to GH support btw).

--- 

After a few rounds of feedback and ideas from everyone involved (thank you all), I think we're getting to the right balance in this API.

This version offers:

- `counter.destroy()`  (and same for all other types)
  - which makes is user friendly and discoverable
  - which lives "right next to `Counter.init`, which is good since both are "lifecycle",
  - and both "reach to MetricsSystem" -- the global store, so it is good to have them close to one another
- `destroyCounter` and friends on the `MetricsFactory` which is the _right place™_ for those functions, since only people _implementing_ the API need to care about what they do with this signal
  - opt-out _is explicit_ so when checking a lib you can quickly see "oh it does nothing on destroy()", according to feedback from previous rounds
  - actual implementation of "releasing resources" is _only_ in the factory thanks to this design
- no way to mistakenly "pass wrong label to delete" (full typesafety for users)
- we stick to explicit lifecycle rather than ARC as there is strong precedence for it, and also we see how ARC lifecycle does not necessarily match the "metric lifecycle" (see writeup in #2)

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -29,6 +29,7 @@ public protocol CounterHandler: AnyObject {
     /// - parameters:
     ///     - by: Amount to increment by.
     func increment(by: Int64)
+
     /// Reset the counter back to zero.
     func reset()
 }
@@ -77,6 +78,7 @@ public class Counter {
     public func reset() {
         self.handler.reset()
     }
+
 }
 
 public extension Counter {
@@ -88,6 +90,13 @@ public extension Counter {
     convenience init(label: String, dimensions: [(String, String)] = []) {
         let handler = MetricsSystem.factory.makeCounter(label: label, dimensions: dimensions)
         self.init(label: label, dimensions: dimensions, handler: handler)
+    }
+
+    /// Signal the underlying metrics library that this counter will never be updated again.
+    /// In response library MAY decide to eagerly release any resources held by this counter.
+    @inlinable
+    func destroy() {
+        MetricsSystem.factory.destroyCounter(self.handler)
     }
 }
 
@@ -168,6 +177,13 @@ public extension Recorder {
     convenience init(label: String, dimensions: [(String, String)] = [], aggregate: Bool = true) {
         let handler = MetricsSystem.factory.makeRecorder(label: label, dimensions: dimensions, aggregate: aggregate)
         self.init(label: label, dimensions: dimensions, aggregate: aggregate, handler: handler)
+    }
+
+    /// Signal the underlying metrics library that this recorder will never be updated again.
+    /// In response library MAY decide to eagerly release any resources held by this recorder.
+    @inlinable
+    func destroy() {
+        MetricsSystem.factory.destroyRecorder(self.handler)
     }
 }
 
@@ -302,6 +318,13 @@ public extension Timer {
         let handler = MetricsSystem.factory.makeTimer(label: label, dimensions: dimensions)
         self.init(label: label, dimensions: dimensions, handler: handler)
     }
+
+    /// Signal the underlying metrics library that this timer will never be updated again.
+    /// In response library MAY decide to eagerly release any resources held by this timer.
+    @inlinable
+    func destroy() {
+        MetricsSystem.factory.destroyTimer(self.handler)
+    }
 }
 
 /// The `MetricsFactory` is the bridge between the `MetricsSystem` and the metrics backend implementation.
@@ -312,6 +335,22 @@ public extension Timer {
 ///
 /// This type is an implementation detail and should not be used directly, unless implementing your own metrics backend.
 /// To use the SwiftMetrics API, please refer to the documentation of `MetricsSystem`.
+///
+/// # Destroying metrics
+///
+/// Since _some_ metrics implementations may need to allocate (potentially "heavy") resources for metrics, destroying
+/// metrics offers a signal to libraries when a metric is "known to never be updated again."
+///
+/// While many metrics are bound to the entire lifetime of an application and thus never need to be destroyed eagerly,
+/// some metrics have well defined unique life-cycles, and it may be beneficial to release any resources held by them
+/// more eagerly than awaiting the applications termination. In such cases, a library or application should invoke
+/// a metric's appropriate `destroy()` method, which in turn results in the corresponding handler that it is backed by
+/// to be passed to `destroyCounter(handler:)`, `destroyRecorder(handler:)` or `destroyTimer(handler:)` where the factory
+/// can decide to free any corresponding resources.
+///
+/// While some libraries may not need to implement this destroying as they may be stateless or similar,
+/// libraries using the metrics API should always assume a library WILL make use of this signal, and shall not
+/// neglect calling these methods when appropriate.
 public protocol MetricsFactory {
     /// Create a backing `CounterHandler`.
     ///
@@ -319,6 +358,7 @@ public protocol MetricsFactory {
     ///     - label: The label for the CounterHandler.
     ///     - dimensions: The dimensions for the CounterHandler.
     func makeCounter(label: String, dimensions: [(String, String)]) -> CounterHandler
+
     /// Create a backing `RecorderHandler`.
     ///
     /// - parameters:
@@ -326,12 +366,34 @@ public protocol MetricsFactory {
     ///     - dimensions: The dimensions for the RecorderHandler.
     ///     - aggregate: Is data aggregation expected.
     func makeRecorder(label: String, dimensions: [(String, String)], aggregate: Bool) -> RecorderHandler
+
     /// Create a backing `TimerHandler`.
     ///
     /// - parameters:
     ///     - label: The label for the TimerHandler.
     ///     - dimensions: The dimensions for the TimerHandler.
     func makeTimer(label: String, dimensions: [(String, String)]) -> TimerHandler
+
+    /// Invoked when the corresponding `Counter`'s `destroy()` function is invoked.
+    /// Upon receiving this signal the factory may eagerly release any resources related to this counter.
+    ///
+    /// - parameters:
+    ///     - handler: The handler to be destroyed.
+    func destroyCounter(_ handler: CounterHandler)
+
+    /// Invoked when the corresponding `Recorder`'s `destroy()` function is invoked.
+    /// Upon receiving this signal the factory may eagerly release any resources related to this recorder.
+    ///
+    /// - parameters:
+    ///     - handler: The handler to be destroyed.
+    func destroyRecorder(_ handler: RecorderHandler)
+
+    /// Invoked when the corresponding `Timer`'s `destroy()` function is invoked.
+    /// Upon receiving this signal the factory may eagerly release any resources related to this timer.
+    ///
+    /// - parameters:
+    ///     - handler: The handler to be destroyed.
+    func destroyTimer(_ handler: TimerHandler)
 }
 
 /// The `MetricsSystem` is a global facility where the default metrics backend implementation (`MetricsFactory`) can be
@@ -363,7 +425,7 @@ public enum MetricsSystem {
         }
     }
 
-    /// Returns a refernece to the configured factory.
+    /// Returns a reference to the configured factory.
     public static var factory: MetricsFactory {
         return self.lock.withReaderLock { self._factory }
     }
@@ -387,6 +449,25 @@ public final class MultiplexMetricsHandler: MetricsFactory {
     public func makeTimer(label: String, dimensions: [(String, String)]) -> TimerHandler {
         return MuxTimer(factories: self.factories, label: label, dimensions: dimensions)
     }
+
+    public func destroyCounter(_ handler: CounterHandler) {
+        for factory in self.factories {
+            factory.destroyCounter(handler)
+        }
+    }
+
+    public func destroyRecorder(_ handler: RecorderHandler) {
+        for factory in self.factories {
+            factory.destroyRecorder(handler)
+        }
+    }
+
+    public func destroyTimer(_ handler: TimerHandler) {
+        for factory in self.factories {
+            factory.destroyTimer(handler)
+        }
+    }
+
 
     private class MuxCounter: CounterHandler {
         let counters: [CounterHandler]
@@ -432,6 +513,7 @@ public final class MultiplexMetricsHandler: MetricsFactory {
 
 /// Ships with the metrics module, used for initial bootstraping.
 public final class NOOPMetricsHandler: MetricsFactory, CounterHandler, RecorderHandler, TimerHandler {
+
     public static let instance = NOOPMetricsHandler()
 
     private init() {}
@@ -439,14 +521,16 @@ public final class NOOPMetricsHandler: MetricsFactory, CounterHandler, RecorderH
     public func makeCounter(label: String, dimensions: [(String, String)]) -> CounterHandler {
         return self
     }
-
     public func makeRecorder(label: String, dimensions: [(String, String)], aggregate: Bool) -> RecorderHandler {
         return self
     }
-
     public func makeTimer(label: String, dimensions: [(String, String)]) -> TimerHandler {
         return self
     }
+
+    public func destroyCounter(_ handler: CounterHandler) {}
+    public func destroyRecorder(_ handler: RecorderHandler) {}
+    public func destroyTimer(_ handler: TimerHandler) {}
 
     public func increment(by: Int64) {}
     public func reset() {}
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /*.xcodeproj
 .xcode
 .SourceKitten
+*.orig
--- a/README.md
+++ b/README.md
@@ -172,6 +172,11 @@ class SimpleMetricsLibrary: MetricsFactory {
     func makeTimer(label: String, dimensions: [(String, String)]) -> TimerHandler {
         return ExampleTimer(label, dimensions)
     }
+    
+    // implementation is stateless, so nothing to do on destroy calls
+    func destroyCounter(_ handler: CounterHandler) {}
+    func destroyRecorder(_ handler: RecorderHandler) {}
+    func destroyTimer(_ handler: TimerHandler) {}
 
     private class ExampleCounter: CounterHandler {
         init(_: String, _: [(String, String)]) {}
--- a/Tests/MetricsTests/CoreMetricsTests+XCTest.swift
+++ b/Tests/MetricsTests/CoreMetricsTests+XCTest.swift
@@ -38,6 +38,9 @@ extension MetricsTests {
             ("testGaugeBlock", testGaugeBlock),
             ("testMUX", testMUX),
             ("testCustomFactory", testCustomFactory),
+            ("testDestroyingGauge", testDestroyingGauge),
+            ("testDestroyingCounter", testDestroyingCounter),
+            ("testDestroyingTimer", testDestroyingTimer),
         ]
     }
 }
--- a/Tests/MetricsTests/CoreMetricsTests.swift
+++ b/Tests/MetricsTests/CoreMetricsTests.swift
@@ -240,4 +240,93 @@ class MetricsTests: XCTestCase {
         let counter2 = Counter(label: "foo", dimensions: [], handler: CustomHandler())
         XCTAssertTrue(counter2.handler is CustomHandler, "expected custom log handler")
     }
+
+    func testDestroyingGauge() throws {
+        let metrics = TestMetrics()
+        MetricsSystem.bootstrapInternal(metrics)
+
+        let name = "gauge-\(NSUUID().uuidString)"
+        let value = Double.random(in: -1000 ... 1000)
+
+        let gauge = Gauge(label: name)
+        gauge.record(value)
+
+        let recorder = gauge.handler as! TestRecorder
+        XCTAssertEqual(recorder.values.count, 1, "expected number of entries to match")
+        XCTAssertEqual(recorder.values.first!.1, value, "expected value to match")
+        XCTAssertEqual(metrics.recorders.count, 1, "recorder should have been stored")
+
+        let identity = ObjectIdentifier(recorder)
+        gauge.destroy()
+        XCTAssertEqual(metrics.recorders.count, 0, "recorder should have been released")
+
+        let gaugeAgain = Gauge(label: name)
+        gaugeAgain.record(-value)
+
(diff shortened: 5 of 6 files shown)


ktoso:
Since I now keep getting 500 errors when trying to view the #3 PR... could I ask someone to close it and direct discussion to here?

Screen Shot 2019-04-16 at 10 29 12

ktoso:
Since I keep getting 500 errors when trying to view the #3 PR, could I ask someone to close that one and direct discussion to this one? Thanks and sorry... Github seems to be freaking out a bit (in touch with GH support about it)...

ktoso:
cc @MrLotU @Mordil (anyone else whom I forgot and can't look up the handles)

ktoso:
Also CC @hartbit, moving discussion here.

I’ll address your comment from the other thread in a moment, grabbing some food 🥘:)

hartbit:
By the way, you spent much more time than me thinking about all this, so I don't want to come around as dismissive. I'm totally ready to agree that `destroy` is the right design. I'm just worried it will be harder to learn for end-users.

Sources/CoreMetrics/Metrics.swift
```diff
@@ -302,6 +318,13 @@ public extension Timer {
         let handler = MetricsSystem.factory.makeTimer(label: label, dimensions: dimensions)
         self.init(label: label, dimensions: dimensions, handler: handler)
     }
+
+    /// Signal the underlying metrics library that this recorder will never be updated again.
```
ianpartridge: s/recorder/timer

weissi:
```suggestion
    /// Signal the underlying metrics library that this timer will never be updated again.
```

weissi: (just trying out this cool GitHub 'suggestion' feature by just plagiarising Ian's suggestion :P )

Sources/CoreMetrics/Metrics.swift
```diff
@@ -432,22 +513,26 @@ public final class MultiplexMetricsHandler: MetricsFactory {
...
-
     public func makeRecorder(label: String, dimensions: [(String, String)], aggregate: Bool) -> RecorderHandler {
         return self
     }
-
     public func makeTimer(label: String, dimensions: [(String, String)]) -> TimerHandler {
         return self
     }
 
+    public func destroy() {}
```
tomerd: left over

ktoso: Whoops thanks, removed

tomerd:
@weissi any further feedback?

weissi:
lgtm

ktoso:
> By the way, you spent much more time than me thinking about all this, so I don't want to come around as dismissive. I'm totally ready to agree that destroy is the right design. I'm just worried it will be harder to learn for end-users.

No worries @hartbit :) That's what review cycles are for 👍 

Copying in the previous concern from PR 3:

> Thanks for the explanation! It's true that ARC does come with design challenges, but I'm also worried about a `destroy` function:
> 
> * Because destroy will sometimes be a no-op, sometimes not (depending on the implementation), 

Correct, though in reality the difference really is about those "short lived" things, which not many people have anyway – asked around some people what they have set up on production, and very rarely is it those "short lived ones." At the same time, I know for certain the ability to do this is very important for some use cases (e.g. we had David chime in about it in https://github.com/apple/swift-metrics/issues/2 and I know of some independent needs for it).

> users might misuse them but think everything is fine (because they are using a backend that is a no-op) but start having issues as soon as they switch backends.

I understand the concern, but also feel that the reality is that application developers are not really completely blind to what they are using. The purpose of this API introduction is to allow calling these things, if we know that "if the backend can support it, I really want it to release now." This matters mostly for external libraries, which don't have any more information about which lib is being used.

> * A destroy function doesn't protect users from continuing to use the metrics objects after being destroyed.

That's true, and how that is treated is up to an implementation. I don't think it is something we need to set in stone in the API here -- some metrics backends have clear semantics around these things but some don't.

> While using ARC might mean abusing it slightly, and cause slightly more headaches in the implementation aspect of things, I does come with greater simplicity for the user.

The worry is not only about that we'd be slightly abusing. We can express an ARC based Counter if we introduce this API, but we can not go the other way around -- if the ARC was the "core api" then we could not _for sure_ implement an explicit release. (You can easily wrap the `Counter` in an `ARCCounter` and call destroy() in release if you _really really_ want to, though I'd argue it's not a good idea). Note also the use case mentioned in #2 which is not possible to implement correctly using ARC -- the "by label, reach for the same gauge and update it." 

Hope this clarifies some more. 
I see we're getting some more LGTMs here so hopefully the main concerns have been addressed :)

tomerd:
@ktoso lmk if you are ready to merge this

ktoso:
We're ready I believe @tomerd !

ktoso:
Oh double checking if readme is up to date...

ktoso:
Hah, of course had forgotten to update readme; fixed and PR should now be now done done™

https://github.com/apple/swift-metrics/pull/17/commits/5d74b3c9b03a4cd102068b3c400eddf6a2e4e1ab

tomerd:
thanks you @ktoso!

MrLotU:
Super awesome! 😄

Issue #5, opened 2019-04-09
overflow


tomerd:
originally brought up by @weissi 

We should think about integer overflows, kinda likely to happen with metrics and you wouldn’t want to crash for metrics.

We could check for them explicitly or use &+ and friends instead of + but that’s only defined behaviour for unsigned ints I think (that’s how it is in C , overflow defined for unsigned and undefined for signed ints).

This is just to remind us to think about it and write tests.

Nio has examples for overflow tests: https://github.com/apple/swift-nio/blob/06649bb8c704a042fc07f2013ae429e2d646e7bb/Tests/NIOTests/ByteBufferTest.swift#L938

milseman:
Which is the lesser of evils here? On wrapping, is it worse to have a counter/metric that goes negative/zero than a trap? Could that lead to a loss of data or a more serious security violation?

weissi:
> Which is the lesser of evils here? On wrapping, is it worse to have a counter/metric that goes negative/zero than a trap? Could that lead to a loss of data or a more serious security violation?

Wrap would definitely be the lesser evil. Metrics are usually egress only, so the emitting program usually doesn't depend on them at all, it shouldn't have any visible effects on the program itself. So crashing for metrics would be pretty bad I think.

weissi:
one question is though if we should maybe send the metrics backend a notification 'by the way, this just overflowed' or so

Lukasa:
Perhaps that notification should log. :wink:

weissi:
> Perhaps that notification should log. 😉

And increase the 'overflow counter metric' haha.

tomerd:
etoomanydependencies

tomerd:
joking aside, i guess we should choose between 
a. capping at max
b. resetting to zero

i slightly prefer a, since b could throw off percentiles calculations and related alerts in a more significant way

milseman:
> Metrics are usually egress only, so the emitting program usually doesn't depend on them at all, it shouldn't have any visible effects on the program itself

This suggests a saturating add, as @tomerd prefers

weissi:
very happy with capping at .max

milseman:
> * cap at max: 👍

The theoretical downside of this approach would be in code that assumed that increment always had some observable effect.

MrLotU:
> The theoretical downside of this approach would be in code that assumed that increment always had some observable effect.

[...]

PR #19, opened 2019-04-18
overflow

tomerd:
motivation: highlight the need to address overflow in metric objects that are integer based

changes:
* add section in the readme about addressing woverflow
* fix timer's unit conversion to address potential overflow

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -243,7 +243,12 @@ public class Timer {
     ///     - value: Duration to record.
     @inlinable
     public func recordMicroseconds<DataType: BinaryInteger>(_ duration: DataType) {
-        self.recordNanoseconds(Int64(duration * 1000))
+        let result = Int64(duration).multipliedReportingOverflow(by: 1000)
+        if result.overflow {
+            self.recordNanoseconds(Int64.max)
+        } else {
+            self.recordNanoseconds(result.partialValue)
+        }
     }
 
     /// Record a duration in microseconds.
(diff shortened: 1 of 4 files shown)


Sources/CoreMetrics/Metrics.swift
```diff
@@ -279,7 +289,12 @@ public class Timer {
     ///     - value: Duration to record.
     @inlinable
     public func recordSeconds<DataType: BinaryInteger>(_ duration: DataType) {
-        self.recordNanoseconds(Int64(duration * 1_000_000_000))
+        let result = Int64(duration).multipliedReportingOverflow(by: 1_000_000_000)
+        if result.overflow {
+            self.recordNanoseconds(Int64.max)
+        } else {
+            self.recordNanoseconds(result.partialValue)
+        }
```
tomerd: capping these to Int64.max is a bit interesting, but i don't see much of an option unless we change the underlying storage type to Double

weissi: `Int64` will hold almost 300 years worth of nanoseconds. I think that's good enough :)

weissi: 2^63/(1_000_000_000*3600*24*365) = 292.4....

ktoso: Yeah, the only case where hitting max happens in such counters is when many many simultanious updates of large values happen, for most metrics its a non problem I think.

weissi: if we believe we need bigger numbers, maybe micro-second resolution is good enough? Then we'd be up to 292,000 years :)

tomerd: y i think we are fine, just wanted to highlight it that its strange in the sense that it can cap the numbers dramatically

tomerd: also, do you feel we should print a warning in these cases?

tomerd:
closes #5

Sources/CoreMetrics/Metrics.swift
ktoso: I don’t think we have to do anything more in the API, an impl could decide to do “the right thing” if it detects the cap is reached. Metrics backends often have a concept of events as well, so they could emit that for example. It also needs state to avoid logging the warning every time the saturated metric is being hit. Best to leave it to impls i think

PR #22, opened 2019-04-23
add script to generate podspec

tomerd:
motivation: distribute the library via podspec

changes:
* add build_podspec.sh script
* adjust sanity check to support more types of copyright years format

PR #23, opened 2019-04-24
Remove NIO headers added in #22

MrLotU:
#22 added double headers (metrics & NIO) to some files. This PR removes them

### DIFF
--- a/scripts/build_podspec.sh
+++ b/scripts/build_podspec.sh
@@ -13,20 +13,6 @@
 ##
 ##===----------------------------------------------------------------------===##
 
-##===----------------------------------------------------------------------===##
-##
-## This source file is part of the SwiftNIO open source project
-##
-## Copyright (c) 2017-2018 Apple Inc. and the SwiftNIO project authors
-## Licensed under Apache License v2.0
-##
-## See LICENSE.txt for license information
-## See CONTRIBUTORS.txt for the list of SwiftNIO project authors
-##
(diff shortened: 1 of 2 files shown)


MrLotU:
Yeah, but these files are not part of the NIO project, right?

weissi:
> Yeah, but these files are not part of the NIO project, right?

They are copied from NIO, so they're now part of both the NIO and SwiftLog projects. That at least is my understanding.

tomerd:
afaik under apache you should preserve the original license header on derived work. http://www.apache.org/legal/src-headers.html#3party

Yasumoto:
Driveby (so feel free to ignore) but it [also mentions](http://www.apache.org/legal/src-headers.html#3party):

> 3. Do not add the standard Apache License header to the top of third-party source files.
> 4. Minor modifications/additions to third-party source files should typically be licensed under the same terms as the rest of the rest of the third-party source for convenience.

@davelester any thoughts? Just curious in general at this point.

davelester:
+1 to @tomerd's comment. If the unmodified files are from another project, the original license headers would apply.

Yasumoto:
Roger that, sounds like we should close this out then?

ktoso:
Also agreed that that’s how copying files from other projects works under Apache.

Thanks a lot for the PR but seems we should close, thanks again @MrLotU :)

Issue #26, opened 2019-05-20
Clarify more visibly in docs intended use SPI/API difference


ktoso:
### Expected behavior

Users should not be confused about how to use the API.

### Actual behavior

In docs and browsing the code the "init that should not be used" is the first one people encounter -- as it is defined on the classes, e.g.:

```swift
    public init(label: String, dimensions: [(String, String)], handler: TimerHandler) {
```

while the one that should be used is only defined later on in an extension:

```swift
    convenience init(label: String, dimensions: [(String, String)] = []) {
```

We should make this easier to spot and understand.

---

There's a few "levels" how much we want to make the spi/api split explicit, let's start with more docs and formatting and we'll see...

ktoso:
Resolved by https://github.com/apple/swift-metrics/pull/27 though will get another follow up PR now

tomerd:
@ktoso do you feel this is now addressed, or needs more work?

ktoso:
It's been a while; let's close this as we've not seen confusion/trouble so far.

PR #27, opened 2019-05-20
doc: #26 clarify which init to use, reformat docs for easier noticing this

ktoso:
### Motivation:

I noticed on a community chat that it was confusing to some that the first init they noticed was documented as "dont use this".

The specific confusion was:

- first found `init(label: String, dimensions: [(String, String)], handler: TimerHandler)` but it has docs "do not use" (correctly, as the intended one to be used is `init(label:dimensions:)`)
- next looked at the `MetricsFactory`, since seems like that's what "creates things"; but this yields the internal handlers; 

Problem: we did not lead the user to the "right" init function to use, as it is defined a bit below -- in an extension, as we wanted it there to be next to the destroy as then all lifecycle for users is in there.

### Modifications:

- adds more see-also links
- reformats main wording that the "this is main api" etc are more visible 

### Result:

Screen Shot 2019-05-20 at 12 10 03

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -36,17 +36,22 @@ public protocol CounterHandler: AnyObject {
 
 /// A counter is a cumulative metric that represents a single monotonically increasing counter whose value can only increase or be reset to zero.
 /// For example, you can use a counter to represent the number of requests served, tasks completed, or errors.
-/// This is the user facing Counter API. Its behavior depends on the `CounterHandler` implementation.
+///
+/// This is the user facing Counter API.
+///
+/// Its behavior depends on the `CounterHandler` implementation.
 public class Counter {
     @usableFromInline
     var handler: CounterHandler
     public let label: String
     public let dimensions: [(String, String)]
 
-    /// Create a new `Counter`.
+    /// Alternative way to create a new `Counter`, while providing an explicit `CounterHandler`.
     ///
     /// This initializer provides an escape hatch for situations where one must use a custom factory instead of the global one.
     /// We do not expect this API to be used in normal circumstances, so if you find yourself using it make sure it's for a good reason.
+    ///
+    /// - SeeAlso: Use `init(label:dimensions:)` to create `Counter` instances using the configured metrics backend.
     ///
     /// - parameters:
     ///     - label: The label for the `Counter`.
@@ -124,7 +129,10 @@ public protocol RecorderHandler: AnyObject {
 }
 
 /// A recorder collects observations within a time window (usually things like response sizes) and *can* provide aggregated information about the data sample, for example, count, sum, min, max and various quantiles.
-/// This is the user facing Recorder API. Its behavior depends on the `RecorderHandler` implementation.
+///
+/// This is the user facing Recorder API.
+///
+/// Its behavior depends on the `RecorderHandler` implementation.
 public class Recorder {
     @usableFromInline
     var handler: RecorderHandler
@@ -132,10 +140,12 @@ public class Recorder {
     public let dimensions: [(String, String)]
     public let aggregate: Bool
 
-    /// Create a new `Recorder`.
+    /// Alternative way to create a new `Recorder`, while providing an explicit `RecorderHandler`.
     ///
     /// This initializer provides an escape hatch for situations where one must use a custom factory instead of the global one.
     /// We do not expect this API to be used in normal circumstances, so if you find yourself using it make sure it's for a good reason.
+    ///
+    /// - SeeAlso: Use `init(label:dimensions:)` to create `Recorder` instances using the configured metrics backend.
     ///
     /// - parameters:
     ///     - label: The label for the `Recorder`.
@@ -221,17 +231,22 @@ public protocol TimerHandler: AnyObject {
 
 /// A timer collects observations within a time window (usually things like request durations) and provides aggregated information about the data sample,
 /// for example, min, max and various quantiles. It is similar to a `Recorder` but specialized for values that represent durations.
-/// This is the user facing Timer API. Its behavior depends on the `TimerHandler` implementation.
+///
+/// This is the user facing Timer API.
+///
+/// Its behavior depends on the `TimerHandler` implementation.
 public class Timer {
     @usableFromInline
     var handler: TimerHandler
     public let label: String
     public let dimensions: [(String, String)]
 
-    /// Create a new `Timer`.
+    /// Alternative way to create a new `Timer`, while providing an explicit `TimerHandler`.
     ///
     /// This initializer provides an escape hatch for situations where one must use a custom factory instead of the global one.
     /// We do not expect this API to be used in normal circumstances, so if you find yourself using it make sure it's for a good reason.
+    ///
+    /// - SeeAlso: Use `init(label:dimensions:)` to create `Recorder` instances using the configured metrics backend.
     ///
     /// - parameters:
     ///     - label: The label for the `Timer`.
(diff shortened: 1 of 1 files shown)


ktoso:
Resolves #26

--- 

We could push it one level more... by separating things into 2 files, where the handlers and factory are in "MetricsSPI.swift" though I know there were some reservations about using the SPI/API wording in docs.

Sources/CoreMetrics/Metrics.swift
```diff
@@ -347,8 +362,8 @@ public extension Timer {
 /// * `Recorder` -> `RecorderHandler`
 /// * `Timer` -> `TimerHandler`
 ///
-/// This type is an implementation detail and should not be used directly, unless implementing your own metrics backend.
-/// To use the SwiftMetrics API, please refer to the documentation of `MetricsSystem`.
+/// - warning: This type is an implementation detail and should not be used directly, unless implementing your own metrics backend.
```
ktoso: I wonder if also all "We do not expect this API to be used in normal circumstances, so if you find yourself using it make sure it's for a good reason." should be made a warning...?

Sources/CoreMetrics/Metrics.swift
```diff
@@ -221,18 +231,23 @@ public protocol TimerHandler: AnyObject {
...
+///
+/// Its behavior depends on the `TimerHandler` implementation.
 public class Timer {
     @usableFromInline
     var handler: TimerHandler
     public let label: String
     public let dimensions: [(String, String)]
 
-    /// Create a new `Timer`.
+    /// Alternative way to create a new `Timer`, while providing an explicit `TimerHandler`.
```
jrtibbetts: Maybe just

    /// Create a new `Timer` with an explicit `TimerHandler`.

Sources/CoreMetrics/Metrics.swift
```diff
@@ -221,18 +231,23 @@ public protocol TimerHandler: AnyObject {
 
 /// A timer collects observations within a time window (usually things like request durations) and provides aggregated information about the data sample,
 /// for example, min, max and various quantiles. It is similar to a `Recorder` but specialized for values that represent durations.
-/// This is the user facing Timer API. Its behavior depends on the `TimerHandler` implementation.
+///
+/// This is the user facing Timer API.
```
jrtibbetts: `user-facing`

Sources/CoreMetrics/Metrics.swift
```diff
@@ -124,19 +129,24 @@ public protocol RecorderHandler: AnyObject {
 }
 
 /// A recorder collects observations within a time window (usually things like response sizes) and *can* provide aggregated information about the data sample, for example, count, sum, min, max and various quantiles.
-/// This is the user facing Recorder API. Its behavior depends on the `RecorderHandler` implementation.
+///
+/// This is the user facing Recorder API.
```
jrtibbetts: `user-facing`

Sources/CoreMetrics/Metrics.swift
```diff
@@ -124,19 +129,24 @@ public protocol RecorderHandler: AnyObject {
...
+/// Its behavior depends on the `RecorderHandler` implementation.
 public class Recorder {
     @usableFromInline
     var handler: RecorderHandler
     public let label: String
     public let dimensions: [(String, String)]
     public let aggregate: Bool
 
-    /// Create a new `Recorder`.
+    /// Alternative way to create a new `Recorder`, while providing an explicit `RecorderHandler`.
```
jrtibbetts: Maybe just

    /// Create a new `Recorder` with an explicit `RecorderHandler`.

jrtibbetts:
It may also help to pull the convenience initializers (i.e. the ones that you want the user to use) out of the extension and put them in the main class body.

Sources/CoreMetrics/Metrics.swift
ktoso: Maybe, though trying to stress the "this is not the api you're looking for" part by using the `Alternative...` spelling 🤔

Sources/CoreMetrics/Metrics.swift
ktoso: Thanks will fix

ktoso:
> It may also help to pull the convenience initializers (i.e. the ones that you want the user to use) out of the extension and put them in the main class body.

Yeah maybe, we should perhaps move all the things then -- including the `destroy()` as the reason for those to be in the extension was for the reason to be "the init and destroy are grouped together there", though at the downside of the more important one being on the class itself and the others on the extension 🤔 

I wonder what @weissi or @tomerd think here

Sources/CoreMetrics/Metrics.swift
tomerd: god idea +1

ktoso: ok will do 👍

ktoso:
Cool, will do the small fixup in the morning then :) Thanks!

Sources/CoreMetrics/Metrics.swift
jrtibbetts: But not everyone reads or notices the function comments, either during autocompletion or by option-clicking the name.

weissi:
> Yeah maybe, we should perhaps move all the things then -- including the `destroy()` as the reason for those to be in the extension was for the reason to be "the init and destroy are grouped together there", though at the downside of the more important one being on the class itself and the others on the extension 🤔

Hmm, how does it help to pull in the convenience initialisers?

Sources/CoreMetrics/Metrics.swift
ktoso: This changes the comment from one wording to another, so not sure what "people dont read comments" aims to highlight here; What's your suggestion?

ktoso:
> Hmm, how does it help to pull in the convenience initialisers?

Only by means of source proximity "first thing when reading top down" could be the intended to be used one; having that said, I don't really think it changes much so I'd stay with the comments changes as this PR does right now.

ktoso:
Added the follow up https://github.com/apple/swift-metrics/pull/27/commits/ff86d5b65efdc8f2c6ae5c7e7019b99cf8e36870

Should be good to go,

weissi:
> > Hmm, how does it help to pull in the convenience initialisers?
> 
> Only by means of source proximity "first thing when reading top down" could be the intended to be used one; having that said, I don't really think it changes much so I'd stay with the comments changes as this PR does right now.

I believe the best way to organise Swift data types is

```
// only the absolutely necessary stuff
struct DataType {
    var actualStoredProperty
    ...
}

// public API
extension DataType {
    public func iCanAlsoDoThis() { ... }
}

// conformance to Fooable
extension DataType: Fooable {
    ...
}
```

because then it can easily be seen what are the stored properties (important to judge the 'width' of a type and its implementation) are. Also it's pretty clear what the public API is (one extension block) and its conformances (other extension blocks). Example [here](https://github.com/apple/swift-nio/pull/1005/files).

Pulling things _into_ the actual type declaration does the opposite. If it's purely a top-down issue, then maybe organise it this way?

```
// public API
extension DataType {
    public func thisIsTheMethodYouShouldUse() {
    }
}


struct DataType {
}
```

ie. put the public API extension first?

tomerd:
to manage scope i suggest we merge this, then address the order / structure in follow up PR if there is consensus it needs to change?

ktoso:
I like the suggestion @weissi 👍 
Can do in a follow up, so this one would only be the docs and I'll do another one with the order changes:

- extension _first_ and it would be the public API really
  - the "right" init and the destroy(),
  - and other funcs as well I guess, like the recordNanoseconds etc
- then class

tomerd:
@ktoso ready to merge this?

ktoso:
Ah sorry about delay here, yes let’s merge and I’ll do the follow up PR shortly

PR #29, opened 2019-05-25
Add guards for Int64 overflow in `Timer` methods

Mordil:
Motivation:

`Timer.record*` methods are generic to cover all cases of `BinaryInteger` values and converts the passed value into `Int64` before doing time unit conversions and overflow checks.

Modifications:

All methods that accept `BinaryInteger` types and converts to `Int64` now guards against values higher than `Int64.max`

Result:

Users can reliably use the `Timer.record*` APIs with `BinaryInteger` types without concern for unknowing fatal errors.

_[One line description of your change]_

### Motivation:

_[Explain here the context, and why you're making that change. What is the problem you're trying to solve.]_

### Modifications:

_[Describe the modifications you've done.]_

### Result:

_[After your change, what will change.]_


tomerd:

tomerd:

PR #31, opened 2019-05-25
Docs #26 shuffle around definitions to show user API first

ktoso:
### Motivation:

It was not entirely clear to people new to the package what the intended init is to call -- as the 'internal backdoor one' was the first init seen when opening the file.

It was also not very clear which types are to be used by users, and which by metrics library authors.

### Modifications:

Following up on the suggestions made on: https://github.com/apple/swift-metrics/pull/27#issuecomment-494298697

- move the extension _first_ such that the lifecycle funcs are the first thing someone sees when browsing the metrics file, and only then the other init etc.
- MARKed the section of this code as `User API`
- moved the handler and factory types to the bottom and under a `MARK: Library SPI`

### Result:

Better separation and order for people consuming the library. Hopefully better docs as well.

Code is roughly now:

```swift
// MARK: User API

public extension Counter {
    convenience init(label: String, dimensions: [(String, String)] = [])
    func destroy()
}
public class Counter { ... }

public extension Recorder {
    convenience init(label: String, dimensions: [(String, String)] = [])
    func destroy()
}
public class Recorder { ... }

public class Gauge: Recorder {
    convenience init(label: String, dimensions: [(String, String)] = [])
    // inherited    func destroy()
}

public extension Timer {
    convenience init(label: String, dimensions: [(String, String)] = [])
    func destroy()
}
public class Timer { ... }

public enum MetricsSystem { ... }

// MARK: Library SPI, intended to be implemented by backend libraries

public protocol MetricsFactory {}

public protocol CounterHandler: AnyObject {}
public protocol RecorderHandler: AnyObject {}
public protocol TimerHandler: AnyObject 

public final class MultiplexMetricsHandler: MetricsFactory {}
public final class NOOPMetricsHandler: MetricsFactory, CounterHandler, RecorderHandler, TimerHandler {}
```

The sections also help in generated jazzy doc: 

Screen Shot 2019-05-25 at 17 49 14
Screen Shot 2019-05-25 at 17 49 10
Screen Shot 2019-05-25 at 17 49 03

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -12,26 +12,25 @@
 //
 //===----------------------------------------------------------------------===//
 
-/// A `CounterHandler` represents a backend implementation of a `Counter`.
-///
-/// This type is an implementation detail and should not be used directly, unless implementing your own metrics backend.
-/// To use the SwiftMetrics API, please refer to the documentation of `Counter`.
-///
-/// # Implementation requirements
-///
-/// To implement your own `CounterHandler` you should respect a few requirements that are necessary so applications work
-/// as expected regardless of the selected `CounterHandler` implementation.
-///
-/// - The `CounterHandler` must be a `class`.
-public protocol CounterHandler: AnyObject {
-    /// Increment the counter.
+// MARK: User API
+
+extension Counter {
+    /// Create a new `Counter`.
     ///
     /// - parameters:
-    ///     - by: Amount to increment by.
-    func increment(by: Int64)
+    ///     - label: The label for the `Counter`.
+    ///     - dimensions: The dimensions for the `Counter`.
+    public convenience init(label: String, dimensions: [(String, String)] = []) {
+        let handler = MetricsSystem.factory.makeCounter(label: label, dimensions: dimensions)
+        self.init(label: label, dimensions: dimensions, handler: handler)
+    }
 
-    /// Reset the counter back to zero.
-    func reset()
+    /// Signal the underlying metrics library that this counter will never be updated again.
+    /// In response the library MAY decide to eagerly release any resources held by this `Counter`.
+    @inlinable
+    public func destroy() {
+        MetricsSystem.factory.destroyCounter(self.handler)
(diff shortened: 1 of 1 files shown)


ktoso:
Resolves #26

Sources/CoreMetrics/Metrics.swift
```diff
@@ -12,26 +12,25 @@
...
-///
-/// To implement your own `CounterHandler` you should respect a few requirements that are necessary so applications work
-/// as expected regardless of the selected `CounterHandler` implementation.
-///
-/// - The `CounterHandler` must be a `class`.
-public protocol CounterHandler: AnyObject {
-    /// Increment the counter.
+// MARK: User API
+
+public extension Counter {
```
weissi: please don't use modifiers in front of `extension`. This modifier propagates to _all_ declarations inside of the extension. In a code review it then frequently becomes impossible to tell if an `internal` or a `public func` has been added. Just remove the `public` here and put it in front of all methods/vars.

ktoso: Okey will do

ktoso: Whoops missed to address this for a long time :O Fixing now

ktoso:
Forgot to follow up here before, done now.

Please have a look @weissi, merge at will? :)

weissi:
@ktoso happy for you to merge but if you wouldn't mind, could you check that this deffo doesn't break public API? The command line (assuming you have a SwiftNIO checkout) should be

    xcrun --toolchain org.swift.5120190702a ~/devel/swift-nio/scripts/check_no_api_breakages.sh ~/devel/swift-metrics SHA_OF_YOUR_COMMIT 1.1.0
    #                          ^^^^                                                                                    ^^^^^
    #                  Swift 5.1 toolchain

ktoso:
Sure I'll check 👍

ktoso:
> OK, all seems good
> done

API checker is happy 👍 

Ship it :shipit:

weissi:
Thanks v much!

Issue #34, opened 2019-07-05
Gauge.record() is under-defined


ktoso:
So I've been using the API a bit, and found an confusing thing:

It is confusing with regards how one should implement `gauge.record()`. An implementation I just reviewed, the prometheus one, implements it by delegating to it's underlying `add` which I think is incorrect, however we don't really provide any guidance about it, so I can see how the author ended up in this:

```swift
// SwiftPrometheus
    func record(_ value: Int64) {
        self.record(value.doubleValue)
    }
    
    func record(_ value: Double) {
        gauge.inc(value, labels)
    }
```

Remember also that we _force_ a gauge to be an _NOT_ aggregated recorder:

```swift
    public convenience init(label: String, dimensions: [(String, String)] = []) {
        self.init(label: label, dimensions: dimensions, aggregate: false)
    }
```

Which again strengthens my idea that it should be a "set" operation.

https://github.com/MrLotU/SwiftPrometheus/blob/master/Sources/PrometheusMetrics/PrometheusMetrics.swift#L37-L42

I think we need to:
- specify which semantics gauge.record -- I think it should be a "set" on the Gauge, and on Histogram it depends I guess but also sounds like a "set"
- I wonder if we are able to offer an add an `add()` really (case for this below)...

----

For reference, the only guidance we provide are:

> /// A recorder collects observations within a time window (usually things like response sizes) and *can* provide aggregated information about the data sample,
> /// for example, count, sum, min, max and various quantiles.

and 

>     /// Record a value.
> `    public func record<DataType: BinaryInteger>(_ value: DataType) {`

Which does not explain if record is a set or add, or if its interpretation should change depending if the recorder / gauge is `aggregate` or not.

----

Note also that open telemetry is converging on gauges offering both and I'd tend to agree with that actually now that I had to use our API a bit:

https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/metrics/GaugeLong.java

- `add`
- `set` -- like our `record`

ktoso:
Paging @tomerd @weissi

tomerd:
thanks @ktoso, as you suggest, the design for recorder and gauge is set semantics. do you think we should change the method name or improve the docs, or something else?

MrLotU:
@ktoso @tomerd I would definitely go with additional documentation, though having both `. add() ` and `.set()` or `.record()` also sounds like a great idea. I’ll at least update my prometheus implementation to use `set()` internally now

[...]

PR #35, opened 2019-07-08
define record() more explicitly that it is a "set the value"

ktoso:
Depends on https://github.com/apple/swift-metrics/pull/31 so please merge that first

Resolves https://github.com/apple/swift-metrics/issues/34

Only commit to look at here is https://github.com/apple/swift-metrics/commit/22dd2161da89e9beca2d99245a4da9cb4f1d1703

The text is a bit dry, but since don't have a "spec" and the types are the spec basically I figured explaining it a bit more could be fine.

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -136,6 +136,7 @@ public class Recorder {
     /// Record a value.
     ///
     /// Recording a value is meant to have "set" semantics, rather than "add" semantics.
+    /// This means that the value of this `Recorder` will match whatever you pass in here.
     /// Metrics implementations may choose to provide additional "add" or other operations meant to modify an existing value,
     /// however the only operation required by this API is the ability to set a specific value.
     ///
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -135,6 +135,10 @@ public class Recorder {
 
     /// Record a value.
     ///
+    /// Recording a value is meant to have "set" semantics, rather than "add" semantics.
+    /// Metrics implementations may choose to provide additional "add" or other operations meant to modify an existing value,
+    /// however the only operation required by this API is the ability to set a specific value.
+    ///
     /// - parameters:
     ///     - value: Value to record.
     @inlinable
@@ -144,6 +148,10 @@ public class Recorder {
 
     /// Record a value.
     ///
+    /// Recording a value is meant to have "set" semantics, rather than "add" semantics.
+    /// Metrics implementations may choose to provide additional "add" or other operations meant to modify an existing value,
+    /// however the only operation required by this API is the ability to set a specific value.
+    /// 
     /// - parameters:
     ///     - value: Value to record.
     @inlinable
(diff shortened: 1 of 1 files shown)


Sources/CoreMetrics/Metrics.swift
```diff
@@ -160,6 +136,10 @@ public class Recorder {
 
     /// Record a value.
     ///
+    /// Recording a value is meant to have "set" semantics, rather than "add" semantics.
```
Yasumoto:
```suggestion
    /// Recording a value is meant to have "set" semantics, rather than "add" semantics.
    /// This means that the value of this `Recorder` will match whatever you pass in here.
```

Took a swing at clarifying `"set" semantics` which I understand, but not sure that's immediately clear without some extra context to all folks.

MrLotU: I personally think just defining `"set"` semantics is clear enough.

tomerd:
@ktoso can you rebase so easier to review the actual change

Sources/CoreMetrics/Metrics.swift
tomerd: this looks good to me

Sources/CoreMetrics/Metrics.swift
```diff
@@ -136,6 +136,10 @@ public class Recorder {
...
+    /// Metrics implementations may choose to provide additional "add" or other operations meant to modify an existing value,
```
tomerd: this seems out of place in the user facing API, maybe move to the Hander API which is aimed for the backend dev?

Sources/CoreMetrics/Metrics.swift
```diff
@@ -145,6 +149,10 @@ public class Recorder {
...
```
tomerd: this seems out of place in the user facing API, maybe move to the Hander API which is aimed for the backend dev?

MrLotU: I think it somewhat makes sense to have this here, for example with Prometheus, if you needed/wanted `add` semantics, you would know that 1) `.record()` has `set` semantics and that 2) you could look into if the underlying `PrometheusClient` does have an `add` operation (which in the case of Prometheus, it does)

tomerd: while i fully recognize the value in using the potentially richer api the "backend" offers, the design goal of metrics-api (and log-api fwiw) is that end users wont need to worry about the underlying "backend". using the "backend" api  is not even an option for libraries that integrate with these apis as they are not exposed to the "backend" which is dictated by the application, but the idea is also true for applications in vast majority of the cases. as such, i think we should not suggest switching to the "backend" api here as it send the opposite message. how about we add a section in the readme that explains this? /cc @ktoso @mrlotu

tomerd: @ktoso @MrLotU wdyt?

ktoso: Whoops seems I missed some emails last week here; Yes, agreed @tomerd 
The API here should not go around and make advertisements for what libraries can't use anyway 👍 

Amending now

ktoso:
Thanks for review everyone, sorry I missed it for a week; 
Addressed your feedback @tomerd, removed the sentence explaining what backends "might" do, agree that wrong place to pitch this: https://github.com/apple/swift-metrics/pull/35/commits/ee94fe5104b7d50001471cc2a6f47359500788d8

PR #42, opened 2019-08-23
TimeUnits

### LINKED ISSUE #39: Timer does not maintain units
### Expected behavior
When I record a duration taking duration as a certain format, I expect the metric to be exposed in that same unit.

### Actual behavior
As described in https://github.com/MrLotU/SwiftPrometheus/issues/10, I'm looking at unexpected behavior where I'm inputting a duration as `Seconds`, but then the unit is exported as nanoseconds.

I understand why we're storing the value as an `Int`, but the unit translation is unexpected when it's being displayed.

This is something that [maybe we could handle in the implementations](https://github.com/MrLotU/SwiftPrometheus/pull/11#pullrequestreview-262818554) but that doesn't feel correct.

### Steps to reproduce

1. Record a duration on a `Timer` using seconds as the unit. `summary.recordSeconds(1.0)`
2. `collect` the metrics.

### If possible, minimal yet complete reproducer code (or URL to code)

https://github.com/MrLotU/SwiftPrometheus/pull/11/files#diff-5b4f613ceb7c7173fa4cddb0c60c9e49R133

### SwiftMetrics version/commit hash

`master`

MrLotU:
Fixes #39 

I created the following Time Units:
* Nano Seconds
* Milli Seconds
* Seconds
* Minutes
* Hours
* Days

It should be quite easy to add/remove from that list but I thought this would be a good start.

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -173,7 +173,7 @@ public class Gauge: Recorder {
 }
 
 public enum TimeUnit {
-    case nanoSeconds, milliSeconds, seconds, minutes, hours, days
+    case nanoseconds, milliseconds, seconds, minutes, hours, days
 }
 
 public extension Timer {
@@ -183,8 +183,8 @@ public extension Timer {
     ///     - label: The label for the `Timer`.
     ///     - displayUnit: A hint to the backend responsible for presenting the data of the preferred display unit. This is not guaranteed to be supported by all backends.
     ///     - dimensions: The dimensions for the `Timer`.
-    convenience init(label: String, prefferedDisplayUnit displayUnit: TimeUnit = .nanoSeconds, dimensions: [(String, String)] = []) {
-        let handler = MetricsSystem.factory.makeTimer(label: label, prefferedDisplayUnit: displayUnit, dimensions: dimensions)
+    convenience init(label: String, preferredDisplayUnit displayUnit: TimeUnit = .nanoseconds, dimensions: [(String, String)] = []) {
+        let handler = MetricsSystem.factory.makeTimer(label: label, preferredDisplayUnit: displayUnit, dimensions: dimensions)
         self.init(label: label, dimensions: dimensions, handler: handler)
     }
 
@@ -402,7 +402,7 @@ public protocol MetricsFactory {
     ///     - label: The label for the `TimerHandler`.
     ///     - displayUnit: A hint to the backend responsible for presenting the data of the preferred display unit. This is not guaranteed to be supported by all backends.
     ///     - dimensions: The dimensions for the `TimerHandler`.
-    func makeTimer(label: String, prefferedDisplayUnit displayUnit: TimeUnit, dimensions: [(String, String)]) -> TimerHandler
+    func makeTimer(label: String, preferredDisplayUnit displayUnit: TimeUnit, dimensions: [(String, String)]) -> TimerHandler
 
     /// Invoked when the corresponding `Counter`'s `destroy()` function is invoked.
     /// Upon receiving this signal the factory may eagerly release any resources related to this counter.
@@ -508,8 +508,8 @@ public final class MultiplexMetricsHandler: MetricsFactory {
         return MuxRecorder(factories: self.factories, label: label, dimensions: dimensions, aggregate: aggregate)
     }
 
-    public func makeTimer(label: String, prefferedDisplayUnit displayUnit: TimeUnit = .nanoSeconds, dimensions: [(String, String)]) -> TimerHandler {
-        return MuxTimer(factories: self.factories, label: label, prefferedDisplayUnit: displayUnit, dimensions: dimensions)
+    public func makeTimer(label: String, preferredDisplayUnit displayUnit: TimeUnit = .nanoseconds, dimensions: [(String, String)]) -> TimerHandler {
+        return MuxTimer(factories: self.factories, label: label, preferredDisplayUnit: displayUnit, dimensions: dimensions)
     }
 
     public func destroyCounter(_ handler: CounterHandler) {
@@ -562,8 +562,8 @@ public final class MultiplexMetricsHandler: MetricsFactory {
 
     private class MuxTimer: TimerHandler {
         let timers: [TimerHandler]
-        public init(factories: [MetricsFactory], label: String, prefferedDisplayUnit displayUnit: TimeUnit = .nanoSeconds, dimensions: [(String, String)]) {
-            self.timers = factories.map { $0.makeTimer(label: label, prefferedDisplayUnit: displayUnit, dimensions: dimensions) }
+        public init(factories: [MetricsFactory], label: String, preferredDisplayUnit displayUnit: TimeUnit = .nanoseconds, dimensions: [(String, String)]) {
+            self.timers = factories.map { $0.makeTimer(label: label, preferredDisplayUnit: displayUnit, dimensions: dimensions) }
         }
 
         func recordNanoseconds(_ duration: Int64) {
@@ -586,7 +586,7 @@ public final class NOOPMetricsHandler: MetricsFactory, CounterHandler, RecorderH
         return self
     }
 
-    public func makeTimer(label: String, prefferedDisplayUnit displayUnit: TimeUnit = .nanoSeconds, dimensions: [(String, String)]) -> TimerHandler {
+    public func makeTimer(label: String, preferredDisplayUnit displayUnit: TimeUnit = .nanoseconds, dimensions: [(String, String)]) -> TimerHandler {
         return self
     }
 
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -181,10 +181,10 @@ public extension Timer {
     ///
     /// - parameters:
     ///     - label: The label for the `Timer`.
-    ///     - storageUnit: The unit to store the time in.
+    ///     - displayUnit: A hint to the backend responsible for presenting the data of the preferred display unit. This is not guaranteed to be supported by all backends.
     ///     - dimensions: The dimensions for the `Timer`.
-    convenience init(label: String, storageUnit: TimeUnit, dimensions: [(String, String)] = []) {
-        let handler = MetricsSystem.factory.makeTimer(label: label, storageUnit: storageUnit, dimensions: dimensions)
+    convenience init(label: String, prefferedDisplayUnit displayUnit: TimeUnit = .nanoSeconds, dimensions: [(String, String)] = []) {
+        let handler = MetricsSystem.factory.makeTimer(label: label, prefferedDisplayUnit: displayUnit, dimensions: dimensions)
         self.init(label: label, dimensions: dimensions, handler: handler)
     }
 
@@ -400,9 +400,9 @@ public protocol MetricsFactory {
     ///
     /// - parameters:
     ///     - label: The label for the `TimerHandler`.
-    ///     - storageUnit: The unit to store the time in.
+    ///     - displayUnit: A hint to the backend responsible for presenting the data of the preferred display unit. This is not guaranteed to be supported by all backends.
     ///     - dimensions: The dimensions for the `TimerHandler`.
-    func makeTimer(label: String, storageUnit: TimeUnit, dimensions: [(String, String)]) -> TimerHandler
+    func makeTimer(label: String, prefferedDisplayUnit displayUnit: TimeUnit, dimensions: [(String, String)]) -> TimerHandler
 
     /// Invoked when the corresponding `Counter`'s `destroy()` function is invoked.
     /// Upon receiving this signal the factory may eagerly release any resources related to this counter.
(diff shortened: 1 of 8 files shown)


Sources/CoreMetrics/Metrics.swift
```diff
@@ -172,14 +172,19 @@ public class Gauge: Recorder {
...
 public extension Timer {
     /// Create a new `Timer`.
     ///
     /// - parameters:
     ///     - label: The label for the `Timer`.
+    ///     - storageUnit: The unit to store the time in.
     ///     - dimensions: The dimensions for the `Timer`.
-    convenience init(label: String, dimensions: [(String, String)] = []) {
-        let handler = MetricsSystem.factory.makeTimer(label: label, dimensions: dimensions)
+    convenience init(label: String, storageUnit: TimeUnit, dimensions: [(String, String)] = []) {
```
tomerd: add default to keep API more backwards compatible

Sources/CoreMetrics/Metrics.swift
```diff
@@ -395,8 +400,9 @@ public protocol MetricsFactory {
     ///
     /// - parameters:
     ///     - label: The label for the `TimerHandler`.
+    ///     - storageUnit: The unit to store the time in.
     ///     - dimensions: The dimensions for the `TimerHandler`.
-    func makeTimer(label: String, dimensions: [(String, String)]) -> TimerHandler
+    func makeTimer(label: String, storageUnit: TimeUnit, dimensions: [(String, String)]) -> TimerHandler
```
tomerd: => add default to keep API more backwards compatible
=> storageUnit => preferredDisplayUnit
=> change the description in the API doc to say: "displayUnit : A hint to the backend responsible for presenting the data of the preferred display unit. This is not guaranteed to be supported by all backends."

Tests/MetricsTests/TestMetrics.swift
```diff
@@ -135,17 +139,33 @@ internal class TestRecorder: RecorderHandler, Equatable {
...
 
-    init(label: String, dimensions: [(String, String)]) {
+    init(label: String, storageUnit: TimeUnit, dimensions: [(String, String)]) {
         self.id = NSUUID().uuidString
         self.label = label
+        self.storageUnit = storageUnit
         self.dimensions = dimensions
     }
 
+    func retrieveValue(atIndex i: Int) -> Int64 {
```
tomerd: move to an extension defined on CoreMetricsTests.swift

Tests/MetricsTests/TestMetrics.swift
```diff
@@ -35,8 +35,12 @@ internal final class TestMetrics: MetricsFactory {
         return self.make(label: label, dimensions: dimensions, registry: &self.recorders, maker: maker)
     }
 
-    public func makeTimer(label: String, dimensions: [(String, String)]) -> TimerHandler {
-        return self.make(label: label, dimensions: dimensions, registry: &self.timers, maker: TestTimer.init)
+    public func makeTimer(label: String, storageUnit: TimeUnit, dimensions: [(String, String)]) -> TimerHandler {
+        self.lock.withLock {
```
tomerd: define a `maker` func like in L32

Sources/CoreMetrics/Metrics.swift
```diff
@@ -172,14 +172,19 @@ public class Gauge: Recorder {
     }
 }
 
+public enum TimeUnit {
+    case nanoSeconds, milliSeconds, seconds, minutes, hours, days
+}
```
tomerd: is there a non-foundation alternative to defining our own units?

/cc @weissi

ktoso: AFAIK no, so sadly currently all projects define their own time/interval etc types.
I think this is better than pulling in foundation for this

weissi: Foundation has

- [the units and measurement stuff](https://developer.apple.com/documentation/foundation/unitduration) which has a bunch of that
- [TimeInterval](https://developer.apple.com/documentation/foundation/timeinterval) which is a `Double` of seconds so can express smaller amounts of time too

Tests/MetricsTests/TestMetrics.swift
MrLotU: Would do, but the lock is inaccessible. I did move the TestCase to `CoreMetricsTests.swift`

Tests/MetricsTests/TestMetrics.swift
MrLotU: I moved away from the maker function here since it does not allow for passing through of the `displayUnit`.

Sources/CoreMetrics/Metrics.swift
MrLotU: I'd love to have a builtin type for this, but I could not find one in my searching.

Tests/MetricsTests/TestMetrics.swift
```diff
@@ -135,17 +139,33 @@ internal class TestRecorder: RecorderHandler, Equatable {
...
     let id: String
     let label: String
+    let displayUnit: TimeUnit
     let dimensions: [(String, String)]
 
     let lock = NSLock()
     var values = [(Date, Int64)]()
 
-    init(label: String, dimensions: [(String, String)]) {
+    init(label: String, prefferedDisplayUnit displayUnit: TimeUnit, dimensions: [(String, String)]) {
```
ktoso: typo? `preffered` -> `preferred` 

https://www.grammar.com/preferred-correct-spelling

Tests/MetricsTests/TestMetrics.swift
ktoso: I see; placing is fine, though naming is a bit misleading; should this be `retriveInPreferredUnit`?

Sources/CoreMetrics/Metrics.swift
ktoso: I guess the doubles argument also has fans... https://twitter.com/ID_AA_Carmack/status/418158611664097280 though not sure about it in the context of metrics.

Tests/MetricsTests/TestMetrics.swift
MrLotU: I will update that!

Tests/MetricsTests/TestMetrics.swift
MrLotU: Aw snap. English really got me like that. I'll fix that! Thanks!

Sources/CoreMetrics/Metrics.swift
```diff
@@ -172,14 +172,19 @@ public class Gauge: Recorder {
     }
 }
 
+public enum TimeUnit {
+    case nanoSeconds, milliSeconds, seconds, minutes, hours, days
```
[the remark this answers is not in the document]
MrLotU: I was in doubt when I wrote these, guess I chose the wrong one 😛 I'll update them.

Sources/CoreMetrics/Metrics.swift
MrLotU: I think for now we can stick with this, since it does not require backends that always use NS to change from seconds to NS now.

Tests/MetricsTests/TestMetrics.swift
MrLotU: I got that, yes. The problem here is that the maker function has to have the signature of `(String, [String, String]) -> Item` so I can not pass through the `preferredDisplayUnit`

Sources/CoreMetrics/Metrics.swift
```diff
@@ -395,8 +400,9 @@ public protocol MetricsFactory {
     ///
     /// - parameters:
     ///     - label: The label for the `TimerHandler`.
+    ///     - displayUnit: A hint to the backend responsible for presenting the data of the preferred display unit. This is not guaranteed to be supported by all backends.
     ///     - dimensions: The dimensions for the `TimerHandler`.
-    func makeTimer(label: String, dimensions: [(String, String)]) -> TimerHandler
+    func makeTimer(label: String, preferredDisplayUnit displayUnit: TimeUnit, dimensions: [(String, String)]) -> TimerHandler
```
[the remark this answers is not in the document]
MrLotU: You can not provide defaults to protocols AFAIK

MrLotU:
I'll update all orders. Thanks for the review @tomerd

tomerd:
@MrLotU  some suggestions ^^

Sources/CoreMetrics/Metrics.swift
```diff
@@ -172,14 +172,19 @@ public class Gauge: Recorder {
...
 public extension Timer {
     /// Create a new `Timer`.
     ///
     /// - parameters:
     ///     - label: The label for the `Timer`.
+    ///     - displayUnit: A hint to the backend responsible for presenting the data of the preferred display unit. This is not guaranteed to be supported by all backends.
     ///     - dimensions: The dimensions for the `Timer`.
-    convenience init(label: String, dimensions: [(String, String)] = []) {
-        let handler = MetricsSystem.factory.makeTimer(label: label, dimensions: dimensions)
+    convenience init(label: String, preferredDisplayUnit displayUnit: TimeUnit = .nanoseconds, dimensions: [(String, String)] = []) {
```
weissi: to not be API breaking, this needs an overload that only has the existing parameters

Sources/CoreMetrics/Metrics.swift
weissi: same here, need an overload with the old parameters to not break API

Sources/CoreMetrics/Metrics.swift
```diff
@@ -502,8 +508,8 @@ public final class MultiplexMetricsHandler: MetricsFactory {
         return MuxRecorder(factories: self.factories, label: label, dimensions: dimensions, aggregate: aggregate)
     }
 
-    public func makeTimer(label: String, dimensions: [(String, String)]) -> TimerHandler {
-        return MuxTimer(factories: self.factories, label: label, dimensions: dimensions)
+    public func makeTimer(label: String, preferredDisplayUnit displayUnit: TimeUnit = .nanoseconds, dimensions: [(String, String)]) -> TimerHandler {
```
weissi: and here and further down

weissi:
@MrLotU could you try running [this script](https://github.com/apple/swift-nio/blob/master/scripts/check_no_api_breakages.sh) over this change. Because currently it breaks the public API, once that checker script is happy, you should be good.

If you need help running it, please reach out.

ktoso:
That reminds me that it'd be a good idea to have a `sanity` script to check API breakage in this repo as we do in NIO perhaps? // @tomerd ?

tomerd:
@ktoso that is only available in 5.1 so need to wait until then

ktoso:
Ah I see... I keep living on the edge and keep forgetting about it... ¯\\_(ツ)_/¯

Tests/MetricsTests/TestMetrics.swift
MrLotU: Ah, yes. My bad. I'll update it!

Sources/CoreMetrics/Metrics.swift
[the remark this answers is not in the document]
MrLotU: This might be a better approach! Massive +1!

tomerd:

Tests/MetricsTests/TestMetrics.swift
```diff
@@ -36,7 +36,10 @@ internal final class TestMetrics: MetricsFactory {
     }
 
     public func makeTimer(label: String, dimensions: [(String, String)]) -> TimerHandler {
-        return self.make(label: label, dimensions: dimensions, registry: &self.timers, maker: TestTimer.init)
+        let maker = { (label: String, dimensions: [(String, String)]) -> TimerHandler in
+            TestTimer(label: label, dimensions: dimensions)
+        }
+        return self.make(label: label, dimensions: dimensions, registry: &self.timers, maker: maker)
```
tomerd: i think this is not required since ctor signature did not change?

tomerd:
also CI is failing since linux tests are not generated, please run `./scripts/generate_linux_tests`

Tests/MetricsTests/TestMetrics.swift
MrLotU: Ah, yes. I'll revert it to what it was before!

MrLotU:
Checking breaking API changes from my `TimeUnits` branch to 41b382a88d3eacabf97345cd9b9e08a82b51c303 gives the following output:
```
==============================
ERROR: public API change in CoreMetrics.json
==============================

/* Generic Signature Changes */

/* RawRepresentable Changes */

/* Removed Decls */

/* Moved Decls */

/* Renamed Decls */

/* Type Changes */

/* Decl Attribute changes */

/* Fixed-layout Type Changes */

/* Protocol Conformance Change */

/* Protocol Requirement Change */
Func TimerHandler.preferDisplayUnit(_:) has been added as a protocol requirement

/* Class Inheritance Change */
Checking Metrics.json... OK
done
```
However I have added a NOOP default implementation for that method in an extension, so this should not qualify as breaking anymore, correct? (CC @weissi)

tomerd:

tomerd:
@MrLotU this looks good but CI fails on formatting, please make sure you run the sanity test locally to make sure everything is ready for merge.

MrLotU:
Will do! Thanks @tomerd

MrLotU:
@tomerd Script is passing locally now! All should be fixed.

tomerd:

tomerd:
@MrLotU i believe one of the unit tests is failing, missing a `return`?

```
func retriveValueInPreferredUnit(atIndex i: Int) -> Int64 {
        return self.lock.withLock {
            let value = values[i].1
            guard let displayUnit = self.displayUnit else {
                return value
            }
            switch displayUnit {
            case .days: return (value / 1_000_000_000) * 60 * 60 * 24
            case .hours: return (value / 1_000_000_000) * 60 * 60
            }
        }
    }
```

MrLotU:
Yup! Saw it too. Pushing right now 😄

tomerd:

tomerd:
@MrLotU https://github.com/apple/swift-metrics/releases/tag/1.20

MrLotU:
Awesome! Thanks

PR #52, opened 2019-11-10
Update Timer.measure() to use DispatchTime instead of Date

MrLotU:
As per @weissi on the Vapor discord:
You’re using wall time which isn’t right to measure how long a function ran. Imagine your CPU went to sleep or the user changed the date.

This came up in a discussion around a similar implementation in SwiftPrometheus
### Motivation:

This PR aims to avoid scenarios where function durations are messed up due to use of wall time.

### Modifications:

Updated `Timer.measure()` to use a delta of `DispatchTime.now().uptimeNanoseconds` instead of wall time.
### Result:

Nothing for end users, although timing will no longer depend on wall time.

### DIFF
--- a/Sources/Metrics/Metrics.swift
+++ b/Sources/Metrics/Metrics.swift
@@ -26,9 +26,10 @@ public extension Timer {
     @inlinable
     static func measure<T>(label: String, dimensions: [(String, String)] = [], body: @escaping () throws -> T) rethrows -> T {
         let timer = Timer(label: label, dimensions: dimensions)
-        let start = Date()
+        let start = DispatchTime.now().uptimeNanoseconds
         defer {
-            timer.record(Date().timeIntervalSince(start))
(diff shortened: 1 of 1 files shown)


weissi:
thanks so much @MrLotU

ktoso:
Hm weird that CI didn't trigger 🤔 

tomerd:
nice. thanks @MrLotU and @weissi

tomerd:
@ktoso latest release is 1.2.0, this would go into 1.2.1 and marked as patch

ktoso:
I see, thanks for fixing! We should close the 1.0 and other already released milestones -- hope you don't mind if I do that @MrLotU @tomerd ?

Closed the milestones that were released: https://github.com/apple/swift-metrics/milestones?state=closed : 1.0.0, 1.1.0, 1.1.1, 1.2.0

PR #54, opened 2019-12-17
better format

tomerd:
motivation: adjust format to xcode format

changes: update .swiftformat no-indent setting and run formatter


.swiftformat
```diff
@@ -8,5 +8,6 @@
 --patternlet inline
 --stripunusedargs unnamed-only
 --comments ignore
+--ifdef no-indent
```
ktoso: Ah yes very good :)

PR #61, opened 2020-02-15
Implement default process & runtime metrics

### LINKED ISSUE #53: Export default process & runtime metrics
From the Prometheus client library spec, `Client libraries SHOULD offer what they can of the Standard exports, documented below.`

Metric name | Help string | Unit
-- | -- | --
process_cpu_seconds_total | Total user and system CPU time spent in seconds. | seconds
process_open_fds | Number of open file descriptors. | file descriptors
process_max_fds | Maximum number of open file descriptors. | file descriptors
process_virtual_memory_bytes | Virtual memory size in bytes. | bytes
process_virtual_memory_max_bytes | Maximum amount of virtual memory available in bytes. | bytes
process_resident_memory_bytes | Resident memory size in bytes. | bytes
process_heap_bytes | Process heap size in bytes. | bytes
process_start_time_seconds | Start time of the process since unix epoch in seconds. | seconds

I think these metrics would be useful to have, also outside of Prometheus.

My proposal is to implement these metrics into `swift-metrics` directly, with the option of switching them off.

Possible API would be:
```swift
MetricsSystem.bootstrap(_ factory: MetricsFactory, provideSystemMetrics: Bool = true)
```

As a note I'd like to add I'm not 100% sure which of the above metrics are feasible and possible to get a hold of in Swift, so some experimentation/research might be needed there.

MrLotU:
Fixes #53 

Currently only implemented on Linux (still figuring out how to handle this on MacOS), and on a 2 second loop.

Points for discussion:
- [ ] Should we move forward without a MacOS implementation since it's _usually_ not used in production environments
- [ ] Should the process metric collection be run on a loop, or be requested at-will by the Metrics Backend
- [ ] Is the 2 second delay between loop runs okay
- [ ] How can we validate these values. Current implementation is copied from [the python library](https://github.com/prometheus/client_python/blob/master/prometheus_client/process_collector.py)


Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,98 @@
...
+                if let rss = Int32(stats[21]) {
+                    Gauge(label: prefix + "resident_memory_bytes").record(rss * Int32(_SC_PAGESIZE))
+                }
+                if let startTimeTicks = Int32(stats[19]) {
+                    Gauge(label: prefix + "start_time_seconds").record(startTimeTicks / ticks)
+                }
+                if let utimeTicks = Int32(stats[11]), let stimeTicks = Int32(stats[12]) {
+                    let utime = utimeTicks / ticks, stime = stimeTicks / ticks
+                    Gauge(label: prefix + "cpu_seconds_total").record(utime + stime)
+                }
```
tomerd: I think we should extract the parsing logic to a separate struct with a ctor that takes the raw string and returns well structured "model"

this also accessed array elements unsafely which could lead to index out of bound crashes, so need to guard against that

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,98 @@
...
+    fileprivate static let queue = DispatchQueue(label: "com.apple.CoreMetrics.SystemMetricsProvider")
+    fileprivate static var shouldRunSystemMetrics: Bool = false
+
+    public static func bootstrapSystemMetrics() {
+        self.lock.withWriterLockVoid {
+            self.shouldRunSystemMetrics = true
+        }
+        DispatchQueue.global(qos: .background).async {
+            updateSystemMetrics()
+        }
```
tomerd: updateSystemMetrics is already async, why use another async context here?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,98 @@
...
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+public enum SystemMetricsProvider {
+    fileprivate static let lock = ReadWriteLock()
+    fileprivate static let queue = DispatchQueue(label: "com.apple.CoreMetrics.SystemMetricsProvider")
+    fileprivate static var shouldRunSystemMetrics: Bool = false
+
+    public static func bootstrapSystemMetrics() {
```
tomerd: would making this an extension of `MetricsSystem` make sense? something like 

`MetricsSystem. monitorSystem()` or  `MetricsSystem.startSystemMonitor()`

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,98 @@
...
+                if !self.shouldRunSystemMetrics {
+                    return true
+                }
+                return false
+            }
+            if shouldReturn { return }
+            let prefix = "process_"
+            let pid = ProcessInfo.processInfo.processIdentifier
+            let ticks = Int32(_SC_CLK_TCK)
+            #if os(Linux)
```
tomerd: I suggest breaking this up to two discrete functions and then calling those depending on operating system. both should return some "model" object that can then be  used to record the metrics consistently across the operating systems

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,98 @@
...
+            self.shouldRunSystemMetrics = false
+        }
+    }
+    
+    private enum SystemMetricsError: Error {
+        case FileNotFound
+    }
+    
+    fileprivate static func updateSystemMetrics() {
+        self.queue.asyncAfter(deadline: .now() + .seconds(2)) {
```
tomerd: the 2 seconds should be configurable (with a default) e.g. `MetricsSystem.startSystemMonitor(pollInterval: TimeUnit = .seconds(2))` etc

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,98 @@
...
+            let shouldReturn = self.lock.withReaderLock { () -> Bool in
+                if !self.shouldRunSystemMetrics {
+                    return true
+                }
+                return false
```
tomerd: consider using a `DispatchQueue::schedule` variant which returns a `Cancellable` so you can cancel on "stop" without managing it yourself

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,98 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+public enum SystemMetricsProvider {
+    fileprivate static let lock = ReadWriteLock()
+    fileprivate static let queue = DispatchQueue(label: "com.apple.CoreMetrics.SystemMetricsProvider")
```
ddossot: Could a `qos` be specified here? Thinking `.background` should do it.

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,98 @@
...
+                else { throw SystemMetricsError.FileNotFound }
+                Gauge(label: prefix + "max_fds").record(maxFds)
+            } catch { print(error) }
+            do {
+                let fm = FileManager.default,
+                    items = try fm.contentsOfDirectory(atPath: "/proc/\(pid)/fd")
+                Gauge(label: prefix + "open_fds").record(items.count)
+            } catch { print(error) }
+            #else
+            #warning("System Metrics are not implemented on MacOS yet.")
```
ddossot: nit: `MacOS` is too specific, it could be another Swift-friendly OS like `iOS`.

Sources/CoreMetrics/SystemMetrics.swift
ddossot: Having two structs - `LinuxOS` and `UnsupportedOS` - would address your suggestion re: "discrete functions" above pretty nicely.

tomerd:
also note the interesting work the perfect folks did in https://github.com/PerfectlySoft/Perfect-SysInfo

Sources/CoreMetrics/SystemMetrics.swift
MrLotU: Reason for this is because I can not access the lock. However if we don't use the `shouldRunSystemMetrics` anymore but use a `Cancellable` like you suggested, that issue would be resolved. Will look into this.

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,98 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Metrics API open source project
+//
+// Copyright (c) 2018-2019 Apple Inc. and the Swift Metrics API project authors
```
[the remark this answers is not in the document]
MrLotU: I think this should be picked up for all files in a separate PR CC @tomerd

Sources/CoreMetrics/SystemMetrics.swift
MrLotU: Addressed all other comments, and seems like we'll still ned a Lock of some sorts. Since the one in `MetricsSystem` is `fileprivate` this will have to be a separate object (or we make the lock non-fileprivate)

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,172 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift Metrics API project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Combine
```
Mordil: Since when could we import this on Linux?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,172 @@
...
+                items = try fm.contentsOfDirectory(atPath: "/proc/\(pid)/fd")
+            self.openFds = Int32(items.count)
+        } catch {
+            self.openFds = nil
+        }
+    }
+}
+
+#else
+private struct UnsuportedMetricsProvider: SystemMetrics {
```
tomerd: should this type be available on all platforms?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,172 @@
...
+// See CONTRIBUTORS.txt for the list of Swift Metrics API project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Combine
+import Foundation
+
+@available(OSX 10.15, *)
```
tomerd: why is this needed?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,172 @@
...
+public enum SystemMetricsProvider {
+    fileprivate static let lock = ReadWriteLock()
+    fileprivate static let queue = DispatchQueue(label: "com.apple.CoreMetrics.SystemMetricsProvider", qos: .background)
+    fileprivate static var timeInterval: DispatchTimeInterval = .seconds(2)
+    fileprivate static var cancellable: Cancellable?
+    fileprivate static var systemMetricsType: SystemMetrics.Type = UnsuportedMetricsProvider.self
+
+    public static func bootstrapSystemMetrics(pollInterval interval: DispatchTimeInterval = .seconds(2), systemMetricsType: SystemMetrics.Type? = nil) {
```
tomerd: I suggest making a self contained struct called SystemMetrics with non-static members and functions. bootstrapSystemMetrics will become its ctor. this will help do away with the locks and state management. then add a static "bootstrapSystemMetrics" method as an extension to MetricsSystem that can only be called once. wdyt?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    public static func cancelSystemMetrics() {
+        self.cancellable?.cancel()
+    }
+
+    fileprivate static func updateSystemMetrics() {
+        let interval = self.lock.withReaderLock { self.timeInterval }
+        let c = self.queue.schedule(after: .init(.now()), interval: .init(interval)) {
+            let prefix = "process_"
```
tomerd: i suggest allowing the users to define a prefix as part of the initial config (like interval)

Sources/CoreMetrics/SystemMetrics.swift
MrLotU: The scheduler that returns a `Cancellable` requires Combine, which in turn needs 10.15, but like @Mordil left in a comment that's not available on Linux. So we probably have to re-think that part entirely. Might be possible to use a `Timer` instead, though not sure how those are with thread safety.

Sources/CoreMetrics/SystemMetrics.swift
ktoso: We can't (-:

Sorry @MrLotU but we'll have to do without Combine. (even if we could, I don't think this library should pull it in)

Sources/CoreMetrics/SystemMetrics.swift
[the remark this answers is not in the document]
ktoso: Technically the rule we are told at  is a more involved: When you touch a file in 2020, you should bump the year to 2020. If a file was not touched in 2020 then it can remain whatever year it was... I don't think many teams follow this strictly in OSS to be honest, but overall that's the latest guidance I've been told 🤔 

Here 2020 would be good in any case I think?

Sources/CoreMetrics/SystemMetrics.swift
ktoso: Yeah, esp since the `_` vs `.` comes into play again 🤔

Sources/CoreMetrics/SystemMetrics.swift
tomerd: how about using DispatchWorkItem or DispatchOperation?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,172 @@
...
+    public static func bootstrapSystemMetrics(pollInterval interval: DispatchTimeInterval = .seconds(2), systemMetricsType: SystemMetrics.Type? = nil) {
+        self.lock.withWriterLockVoid {
+            self.timeInterval = interval
+            if let type = systemMetricsType {
+                self.systemMetricsType = type
+            } else {
+                #if os(Linux)
+                self.systemMetricsType = LinuxSystemMetricsProvider.self
+                #else
+                self.systemMetricsType = UnsuportedMetricsProvider.self
```
ktoso: Rather: 

```suggestion
                self.systemMetricsType = UnsuportedSystemMetricsProvider.self
```

i.e. we're on a system whose system metrics are not implemented

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,172 @@
...
+        let c = self.queue.schedule(after: .init(.now()), interval: .init(interval)) {
+            let prefix = "process_"
+            let pid = ProcessInfo.processInfo.processIdentifier
+            let metrics = self.systemMetricsType.init(pid: "\(pid)")
+            if let vmem = metrics.virtualMemory { Gauge(label: prefix + "virtual_memory_bytes").record(vmem) }
+            if let rss = metrics.residentMemory { Gauge(label: prefix + "resident_memory_bytes").record(rss) }
+            if let start = metrics.startTimeSeconds { Gauge(label: prefix + "start_time_seconds").record(start) }
+            if let cpuSeconds = metrics.cpuSeconds { Gauge(label: prefix + "cpu_seconds_total").record(cpuSeconds) }
+            if let maxFds = metrics.maxFds { Gauge(label: prefix + "max_fds").record(maxFds) }
+            if let openFds = metrics.openFds { Gauge(label: prefix + "open_fds").record(openFds) }
```
ktoso: Should we allow customization of all labels? I.e. make a settings struct where all the labels are vars and allow passing in the settings when creating SystemMetrics instance?

Sources/CoreMetrics/SystemMetrics.swift
ktoso: +1, no need to hide this one.

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,172 @@
...
+private struct UnsuportedMetricsProvider: SystemMetrics {
+    let virtualMemory: Int32?
+    let residentMemory: Int32?
+    let startTimeSeconds: Int32?
+    let cpuSeconds: Int32?
+    let maxFds: Int32?
+    let openFds: Int32?
+
+    init(pid: String) {
+        #warning("System Metrics are not implemented on non-Linux platforms yet.")
```
ktoso: So this worked ok while we hide this type on non linux, but I don't think we should hide it, so this line has to be removed in order to not cause compilation warnings

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,172 @@
...
+
+private extension Array where Element == String {
+    subscript(safe index: Int) -> String {
+        guard index >= 0, index < endIndex else {
+            return ""
+        }
+
+        return self[index]
+    }
+}
```
ktoso: Can we remove this extension and to a guard/if instead?
Let's try to keep the extensions on arrays etc minimal in such projects -- even if private.

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,172 @@
...
+        do {
+            guard let statString =
+                try String(contentsOfFile: "/proc/\(pid)/stat", encoding: .utf8)
+                .split(separator: ")")
+                .last
+            else { throw SystemMetricsError.FileNotFound }
+            let stats = String(statString)
+                .split(separator: " ")
+                .map(String.init)
+            self.virtualMemory = Int32(stats[safe: 20])
```
ktoso: not super keen on introducing one off `safe:` accessors like that, rather let's do the `if let value = ... { self.virtualMemory = value` usual dance?

ktoso: OR `?? ""` when needed?

ktoso: OR a form of `getIntStat(stats, 11)` etc?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,172 @@
...
+
+public protocol SystemMetrics {
+    init(pid: String)
+
+    var virtualMemory: Int32? { get }
+    var residentMemory: Int32? { get }
+    var startTimeSeconds: Int32? { get }
+    var cpuSeconds: Int32? { get }
+    var maxFds: Int32? { get }
+    var openFds: Int32? { get }
```
ktoso: Are we sure about the `?`? I know it's kind of "right" but all those are numerics where `0` can be used as a default if we could not read... WDYT?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,172 @@
...
+}
+
+#if os(Linux)
+private struct LinuxSystemMetricsProvider: SystemMetrics {
```
ktoso: If implementations are called Provider then the protocol should be to -- `SystemMetricsProvider`?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,172 @@
...
+
+    public static func cancelSystemMetrics() {
+        self.cancellable?.cancel()
+    }
+
+    fileprivate static func updateSystemMetrics() {
+        let interval = self.lock.withReaderLock { self.timeInterval }
+        let c = self.queue.schedule(after: .init(.now()), interval: .init(interval)) {
+            let prefix = "process_"
+            let pid = ProcessInfo.processInfo.processIdentifier
```
ktoso: Should we read and store the pid just once rather than many times, is this lookup cheap?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,172 @@
...
+            if let openFds = metrics.openFds { Gauge(label: prefix + "open_fds").record(openFds) }
+        }
+        self.lock.withWriterLockVoid {
+            self.cancellable = c
+        }
+    }
+}
+
+public protocol SystemMetrics {
+    init(pid: String)
```
ktoso: Passing PID as string seems a bit weird, pass as Int and string it inside the impl?

weissi: If we’re doing this, we need to also think about Windows, so paging @compnerd

compnerd: Windows uses a `DWORD` (type alias for Int32).  Id prefer `process` or `id` rather than `pid` since `pid` and `pid_t` are pretty Unix centric.

I completely agree with @ktoso that `String` is an odd type for this.  Why not have something like:

```
#if os(Windows)
typealias ProcessID = DWORD
#else
typealias ProcessID = pid_t
#endif
```

Sources/CoreMetrics/SystemMetrics.swift
MrLotU: The reasoning for optional is the difference between `0` and `nil` for reporting. A metric not being supported and thus not exported is different from that metric being `0`. This could cause issues with (for example) alerting systems. I think in this case it's a better safe than sorry situation. But if I'm alone in this, we can change it 😄

Sources/CoreMetrics/SystemMetrics.swift
MrLotU: Reasoning for this is that normally an array does not return optionals or throws errors. It will crash if you go out of index, thus the `[safe:]` subscript that checks the bounds. I'm okay with moving this to a separate function, but don't fully see the point of that.

Sources/CoreMetrics/SystemMetrics.swift
MrLotU: See https://github.com/apple/swift-metrics/pull/61/files#r398495089 for reasoning on this 😄

Sources/CoreMetrics/SystemMetrics.swift
MrLotU: I have gone with @compnerd 's solution here, but wanted to validate that on windows ` ProcessInfo.processInfo.processIdentifier` will return a `DWORD`, otherwise we need an extra check here: https://github.com/apple/swift-metrics/pull/61/files#diff-b35e7c2396a198ceca87c72cdc27817fR43

MrLotU:
Thanks for all the review comments. I finally got around to updating everything. As suggested by @tomerd I've gone with a change in approach. `SystemMetricsHandler` is now a self-enclosed class that will take care of everything, with `MetricsSystem` handeling as it's ctor. Enabeling system metrics now looks like this:
```swift
MetricsSystem.bootstrapSystemMetrics(
    labels: SystemMetricsLabels(
        prefix: "process_",
        virtualMemory: "virtual_memory_bytes",
        residentMemory: "resident_memory_bytes",
        startTimeSeconds: "start_time_seconds",
        cpuSecondsTotal: "cpu_seconds_total",
        maxFds: "max_fds",
        openFds: "open_fds"
    )
)
```

What I will probably do for `SwiftPrometheus` to reduce user-required copy-pasta is add something like
```swift
public extension MetricsSystem {
    static func bootstrapSystemMetrics() {
        self.bootstrapSystemMetrics(
            labels: SystemMetricsLabels(
                prefix: "process_",
                virtualMemory: "virtual_memory_bytes",
                residentMemory: "resident_memory_bytes",
                startTimeSeconds: "start_time_seconds",
                cpuSecondsTotal: "cpu_seconds_total",
                maxFds: "max_fds",
                openFds: "open_fds"
            )
        )
    }
}
```
allowing users to just `MetricsSystem.bootstrapSystemMetrics()` and I would advice other library owners to do the same.

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,204 @@
...
+// See CONTRIBUTORS.txt for the list of Swift Metrics API project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+#if os(Windows)
+public typealias ProcessId = DWORD
```
weissi: why don't we just make them `Int`? That's the recommendation in the [Swift language guide](https://docs.swift.org/swift-book/LanguageGuide/TheBasics.html#ID324).

> Numeric Type Conversion
>
> Use the Int type for all general-purpose integer constants and variables in your code, even if they’re known to be nonnegative. Using the default integer type in everyday situations means that integer constants and variables are immediately interoperable in your code and will match the inferred type for integer literal values.
>
> Use other integer types only when they’re specifically needed for the task at hand, because of explicitly sized data from an external source, or for performance, memory usage, or other necessary optimization. Using explicitly sized types in these situations helps to catch any accidental value overflows and implicitly documents the nature of the data being used.
>

weissi: should we be worried that `Int` is too small on 32 bit platforms, then we should go with `Int64` IMHO

Sources/CoreMetrics/SystemMetrics.swift
weissi: - They should all be `Int`, no? why don't we just make them `Int` as quoted above, that's what the [Swift language guide](https://docs.swift.org/swift-book/LanguageGuide/TheBasics.html#ID324) recommends.
- the unit is missing in all the fields. This should be `virtualMemoryBytes`, right? (And then `Int32` is also definitely too small)

Sources/CoreMetrics/SystemMetrics.swift
compnerd: > why don't we just make them `Int`? That's the recommendation in the [Swift language guide](https://docs.swift.org/swift-book/LanguageGuide/TheBasics.html#ID324).
> 
> 

The recommendation didn’t work well here.  A DWORD is the system type for this and is not an Int.  Using Int will cause type conversations which need to be made explicit.  I’m strongly against using Int for the process type.


> 
> > Numeric Type Conversion
> 
> >
> 
> > Use the Int type for all general-purpose integer constants and variables in your code, even if they’re known to be nonnegative. Using the default integer type in everyday situations means that integer constants and variables are immediately interoperable in your code and will match the inferred type for integer literal values.
> 
> >
> 
> > Use other integer types only when they’re specifically needed for the task at hand, because of explicitly sized data from an external source, or for performance, memory usage, or other necessary optimization. Using explicitly sized types in these situations helps to catch any accidental value overflows and implicitly documents the nature of the data being used.
> 
> >

weissi: @compnerd / @MrLotU they can't be different on different platforms because it's a `public protocol`. Otherwise you'd get implementations that only confirm on say Linux and not on Windows.

compnerd: They are different across the platforms.  I suppose that crafting a wrapper that is constructed by the platforms type should alleviate your concern.

weissi: wrapper works but I was thinking of `Int` as the "wrapper". `Int(myDWORD)` works, just like `Int(my_pid_t)` does.

Sources/CoreMetrics/Metrics.swift
```diff
@@ -363,6 +365,8 @@ public enum MetricsSystem {
     fileprivate static let lock = ReadWriteLock()
     fileprivate static var _factory: MetricsFactory = NOOPMetricsHandler.instance
     fileprivate static var initialized = false
+    fileprivate static var systemMetricsHandler: SystemMetricsHandler?
+    fileprivate static var systemMetricsInitialized = false
```
tomerd: do we need two flags? isn't systemMetricsHandler being nil a good enough indication?

Sources/CoreMetrics/Metrics.swift
```diff
@@ -378,6 +382,33 @@ public enum MetricsSystem {
...
+    /// `bootstrapSystemMetrics` is an one-time configuration function which globally enables system level metrics.
+    /// `bootstrapSystemMetrics` can be only called once, unless cancelled usung `cancelSystemMetrics`, calling it more
+    /// than once without cancelling will lead to undefined behaviour, most likely a crash.
+    ///
+    /// - parameters:
+    ///     - pollInterval: The interval at which system metrics should be updated.
+    ///     - systemMetricsType: The type of system metrics to use. If none is provided this defaults to
+    ///                          `LinuxSystemMetrics` on Linux platforms and `NOOPSystemMetrics` on all other platforms.
+    ///     - labels: The labels to use for generated system metrics.
+    public static func bootstrapSystemMetrics(pollInterval interval: DispatchTimeInterval = .seconds(2), systemMetricsType: SystemMetrics.Type? = nil, labels: SystemMetricsLabels) {
```
tomerd: the way I imagined this would work is a separate module that extends `MetricsSystem` and adds a new variant of `bootstrap`, something like: 

```
public static func bootstrap(_ factory: MetricsFactory, systemMetrics: SystemMetricsConfiguration) {
```
having  SystemMetricsConfiguration a container for the `pollInternval` and other configuration options

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,198 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift Metrics API project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
```
tomerd: CoreMetrics is designed be foundation "free", see point about separate module ^^

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,198 @@
...
+public protocol SystemMetrics {
+    init(pid: Int)
+
+    var virtualMemoryBytes: Int? { get }
+    var residentMemoryBytes: Int? { get }
+    var startTimeSeconds: Int? { get }
+    var cpuSeconds: Int? { get }
+    var maxFds: Int? { get }
+    var openFds: Int? { get }
+}
```
tomerd: how about we separate the function and the data? ie make `SystemMetrics` a simple struct that just carried the data (model object if you will) and introduce a protocol `SystemMetricsReader`

```
protocol SystemMetricsReader {
  func read(pid: Int) -> Result<SystemMetrics, Error>
}
```

this way we dont need all these optionals and can simplify the try/catch stuff - imo if it fails than no data better than partial data

tomerd:
thanks for updating the PR @MrLotU this is getting closer! left a few suggestions

Sources/CoreMetrics/Metrics.swift
MrLotU: I'd like to do that, but because of the access modifiers in `MetricsSystem` this is not possible. I have managed to lose the `Foundation` dependency, so there might not be a reason anymore to split out?

Sources/CoreMetrics/Metrics.swift
MrLotU: Good point. Will remove the flag

Sources/CoreMetrics/SystemMetrics.swift
MrLotU: Will remove the Foundation dep 👍

Sources/CoreMetrics/SystemMetrics.swift
MrLotU: The reasoning for the optionals here is to always provide as much data as possible to users. For example, on Linux these metrics are retrieved in 3 parts. 1 of those parts could fail, but the other metrics will still be exposed.

Curious what your reasoning is for no data better than partial data. (IMO it's not partial data, just less data)

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,235 @@
...
+    init(pid: Int) {
+        let pid = "\(pid)"
+        let ticks = _SC_CLK_TCK
+        
+        do {
+            guard
+                let fp = fopen("/proc/\(pid)/stat", "r")
+            else { throw SystemMetricsError.MetricReadError }
+            var buf = [CChar](repeating: CChar(0), count: 1024)
+            while fgets(&buf, 1024, fp) != nil { }
```
weissi, line 153: - this isn't right. This overwrites `buf` in every loop iteration.
- can we please use `buf.withUnsafeBytes { ptr in while fgets(...) }` instead of relying on the dangerous, magic conversion from `&` to `Unsafe*Pointer`?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,235 @@
...
+            guard fclose(fp) == 0 else { throw SystemMetricsError.MetricReadError }
+            guard
+                let statString = String(cString: buf)
+                .split(separator: ")")
```
weissi, line 157: this probably isn't right if the process name contains a `)`

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,235 @@
...
+                .last
+            else { throw SystemMetricsError.MetricReadError }
+            let stats = String(statString)
+                .split(separator: " ")
```
weissi, line 161: this probably isn't right if the process name contains a space

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,235 @@
...
+        } catch {
+            self.virtualMemoryBytes = nil
+            self.residentMemoryBytes = nil
+            self.startTimeSeconds = nil
+            self.cpuSeconds = nil
+        }
+        do {
+            var _rlim = rlimit()
+            
+            guard getrlimit(__rlimit_resource_t(RLIMIT_NOFILE.rawValue), &_rlim) == 0 else { throw SystemMetricsError.MetricReadError }
```
weissi, line 189: here as well, please swap `&_rlim` for `with*Unsafe*(&_rlim) { rLimPtr in ... }`

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,235 @@
...
+            self.maxFds = Int(_rlim.rlim_max)
+        } catch {
+            self.maxFds = nil
+        }
+        do {
+            guard let dir = opendir("/proc/\(pid)/fd") else { throw SystemMetricsError.MetricReadError }
```
weissi, line 195: why not just use `/proc/self/...` ?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,235 @@
...
+            var count = 0
+            while readdir(dir) != nil { count += 1 }
+            guard closedir(dir) == 0 else { throw SystemMetricsError.MetricReadError }
```
weissi, line 198: can we please match resource allocation (`opendir`) and resource deallocation (`closedir`) this way

    let resource = try allocateResource()
    defer {
        deallocateResource(resource)
   }

in all cases? Otherwise it's just too easy to introduce a `throw` or a `try` between resource allocation & deallocation which will then leak.

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,235 @@
...
+            guard let task = self.task else { return }
+            task.perform()
+            self.updateSystemMetrics()
+        }
+    }
+}
+
+public struct SystemMetricsLabels {
+    let prefix: String
+    let virtualMemoryBytes: String
```
weissi, line 98: can we add a link to the spec that specified that those are the "system metrics" and also the right names? `Fds` sounds overly abbreviated and also UNIX-centric?

Sources/CoreMetrics/SystemMetrics.swift
tomerd: @MrLotU I am interested to understand the failure scenarios. i.e if some of these files you are trying to read does not exist, things must be bad no?

Sources/CoreMetrics/Metrics.swift
tomerd: @MrLotU what about something like: 

```
import CoreMetrics
import Dispatch

extension MetricsSystem {
    private static var systemMetricsFactory: SystemMetricsFactory?
    
    public static func bootstrapWithSystemMetrics(_ factory: MetricsFactory, configuration: SystemMetrics.Configuration) {
        let factory = SystemMetricsFactory(factory: factory, configuration: configuration)
        self.bootstrap(factory)
        systemMetricsFactory = factory
        systemMetricsFactory!.poll()
    }
    
    public static func cancelSystemMetrics() {
        systemMetricsFactory?.cancelPolling()
    }
    
    internal struct SystemMetricsFactory: MetricsFactory {
        let underlying: MetricsFactory
        let configuration: SystemMetrics.Configuration

        let queue = DispatchQueue(label: "com.apple.CoreMetrics.SystemMetricsHandler", qos: .background)
        let task: DispatchWorkItem
        
        init (factory: MetricsFactory, configuration: SystemMetrics.Configuration) {
            self.underlying = factory
            self.configuration = configuration
            self.task = DispatchWorkItem(qos: .background, block: {
                let metrics = SystemMetricsFactory.readSystemMetrics()
                Gauge(label: configuration.labels.label(for: \.virtualMemoryBytes)).record(metrics.virtualMemoryBytes)
                Gauge(label: configuration.labels.label(for: \.residentMemoryBytes)).record(metrics.residentMemoryBytes)
                Gauge(label: configuration.labels.label(for: \.startTimeSeconds)).record(metrics.startTimeSeconds)
                Gauge(label: configuration.labels.label(for: \.cpuSecondsTotal)).record(metrics.cpuSeconds)
                Gauge(label: configuration.labels.label(for: \.maxFds)).record(metrics.maxFds)
                Gauge(label: configuration.labels.label(for: \.openFds)).record(metrics.openFds)
            })
        }
        
        func poll() {
            self.queue.asyncAfter(deadline: .now() + self.configuration.interval) {
                guard !self.task.isCancelled else {
                    return
                }
                self.task.perform()
                self.poll()
            }
        }
        
        func cancelPolling() {
            self.task.cancel()
        }
        
        // MARK: - proxy
        
        func makeCounter(label: String, dimensions: [(String, String)]) -> CounterHandler {
            self.underlying.makeCounter(label: label, dimensions: dimensions)
        }
        
        func makeRecorder(label: String, dimensions: [(String, String)], aggregate: Bool) -> RecorderHandler {
            self.underlying.makeRecorder(label: label, dimensions: dimensions, aggregate: aggregate)
        }
        
        func makeTimer(label: String, dimensions: [(String, String)]) -> TimerHandler {
            self.underlying.makeTimer(label: label, dimensions: dimensions)
        }
        
        func destroyCounter(_ handler: CounterHandler) {
            self.underlying.destroyCounter(handler)
        }
        
        func destroyRecorder(_ handler: RecorderHandler) {
            self.underlying.destroyRecorder(handler)
        }
        
        func destroyTimer(_ handler: TimerHandler) {
            self.underlying.destroyTimer(handler)
        }
        
        // MARK: - system metrics
        
        private static func readSystemMetrics() -> SystemMetrics.Data {
            fatalError("FIXME")
        }
    }
    
}

public enum SystemMetrics {
        
    public struct Configuration {
        let interval: DispatchTimeInterval
        let labels: Labels
        let metricsType: SystemMetrics.Type?
        
        public init(pollInterval interval: DispatchTimeInterval = .seconds(2),
                    labels: Labels,
                    metricsType: SystemMetrics.Type? = nil) {
            self.interval = interval
            self.labels = labels
            self.metricsType = metricsType
            
        }
    }
    
    public struct Labels {
        let prefix: String
        let virtualMemoryBytes: String
        let residentMemoryBytes: String
        let startTimeSeconds: String
        let cpuSecondsTotal: String
        let maxFds: String
        let openFds: String

        public init(prefix: String, virtualMemoryBytes: String, residentMemoryBytes: String, startTimeSeconds: String, cpuSecondsTotal: String, maxFds: String, openFds: String) {
            self.prefix = prefix
            self.virtualMemoryBytes = virtualMemoryBytes
            self.residentMemoryBytes = residentMemoryBytes
            self.startTimeSeconds = startTimeSeconds
            self.cpuSecondsTotal = cpuSecondsTotal
            self.maxFds = maxFds
            self.openFds = openFds
        }

        func label(for keyPath: KeyPath<SystemMetrics.Labels, String>) -> String {
            return self.prefix + self[keyPath: keyPath]
        }
    }
    
    public struct Data {
        let pid: Int
        let virtualMemoryBytes: Int
        let residentMemoryBytes: Int
        let startTimeSeconds: Int
        let cpuSeconds: Int
        let maxFds: Int
        let openFds: Int
    }
}
```

have not tested it but I think can be made to work

tomerd:
het @mrlotu, wdyt about https://github.com/apple/swift-metrics/pull/61#discussion_r411646471

MrLotU:
I've been quite busy over the last week(s) and hadn't had time to look at this yet. I just read your comment and it looks promising! I will take another stab at it hopefully this weekend.

@weissi I did already take an initial look into your comments and am afraid I'm going to need a bit more help on that end. I've tried using the methods you mentioned (`withUnsafeBytes` and the likes) but couldn't get it to compile that way. Pointers (no pun intended) would be appreciated 😄

MrLotU:
@tomerd I've addressed your comments from https://github.com/apple/swift-metrics/pull/61#discussion_r411646471 to the best of my abilities. Would appreciate another round of review/feedback. 😄

Sources/CoreMetrics/Metrics.swift
```diff
@@ -384,9 +386,16 @@ public enum MetricsSystem {
...
+    
+    internal static var systemMetrics: SystemMetricsFactory? {
+        return self.lock.withReaderLock { _factory as? SystemMetricsFactory }
+    }
 
     /// Returns a reference to the configured factory.
     public static var factory: MetricsFactory {
+        if let f = self.systemMetrics {
+            return f.underlying
+        }
```
tomerd, line 398: is this in place to save a hop?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,322 @@
...
+    /// `bootstrapSystemMetrics` can be only called once, unless cancelled usung `cancelSystemMetrics`, calling it more
+    /// than once without cancelling will lead to undefined behaviour, most likely a crash.
+    /// There is no need to call `bootstrapSystemMetrics` if `SystemMetricsOptions` were provided to `bootstrap`.
+    ///
+    /// - parameters:
+    ///     - options: Options to configure `SystemMetricsHandler`
+    public static func bootstrapWithSystemMetrics(_ factory: MetricsFactory, options: SystemMetrics.Options) {
+        let factory = SystemMetricsFactory(factory: factory, options: options)
+        self.bootstrap(factory)
+        self.systemMetrics!.poll()
```
tomerd, line 32: maybe just `factory.poll`?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,322 @@
...
+    }
+    
+    /// Cancels the collection of system metrics. Calling this unlocks `bootstrapSystemMetrics` so it can be
+    /// called again.
+    public static func cancelSystemMetrics() {
```
tomerd, line 37: can you elaborate when you see this API being used

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,322 @@
...
+        self.systemMetrics?.cancelSystemMetrics()
+    }
+    
+    internal class SystemMetricsFactory: MetricsFactory {
+        fileprivate let queue = DispatchQueue(label: "com.apple.CoreMetrics.SystemMetricsHandler", qos: .background)
+        fileprivate let timeInterval: DispatchTimeInterval
+        fileprivate let systemMetricsType: SystemMetricsProvider.Type
+        fileprivate let labels: SystemMetrics.Labels
+        fileprivate var task: DispatchWorkItem?
```
tomerd, line 46: do we need a lock around task?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,322 @@
...
+
+public protocol SystemMetricsProvider {
+    static func readSystemMetrics() -> SystemMetrics.Data?
+}
+
+public enum SystemMetrics {
+    /// Options used to bootstrap `SystemMetricsHandler`
+    ///
+    /// Libraries are advised to extend `SystemMetricsOptions` with a static instance for ease of use.
+    public struct Options {
```
tomerd, line 121: We typically names such `Configuration`

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,322 @@
...
+        }
+        
+        func destroyTimer(_ handler: TimerHandler) {
+            self.underlying.destroyTimer(handler)
+        }
+    }
+}
+
+public protocol SystemMetricsProvider {
+    static func readSystemMetrics() -> SystemMetrics.Data?
```
tomerd, line 114: is there a strong reason to use static func? I'd even argue that this can be simplified to:

`public typealias SystemMetrics.DataProvider = () -> SystemMetrics.Data?`, and then the user can provide a generator of their choice

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,322 @@
...
+        let labels: Labels
+        
+        /// Create new instance of `SystemMetricsOptions`
+        ///
+        /// - parameters:
+        ///     - pollInterval: The interval at which system metrics should be updated.
+        ///     - systemMetricsType: The type of system metrics to use. If none is provided this defaults to
+        ///                          `LinuxSystemMetrics` on Linux platforms and `NOOPSystemMetrics` on all other platforms.
+        ///     - labels: The labels to use for generated system metrics.
+        public init(pollInterval interval: DispatchTimeInterval = .seconds(2), metricsType: SystemMetricsProvider.Type? = nil, labels: Labels) {
```
tomerd, line 133: see comment above about changing  SystemMetricsProvider.Type to a functional generator

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,322 @@
...
+            let maxFileDescriptors = Int(_rlim.rlim_max)
+
+            guard let dir = opendir("/proc/self/fd") else { return nil }
+            defer {
+                closedir(dir)
+            }
+            var openFileDescriptors = 0
+            while readdir(dir) != nil { openFileDescriptors += 1 }
+            
+            return .init(virtualMemoryBytes: virtualMemoryBytes, residentMemoryBytes: residentMemoryBytes, startTimeSeconds: startTimeSeconds, cpuSeconds: cpuSeconds, maxFileDescriptors: maxFileDescriptors, openFileDescriptors: openFileDescriptors)
```
tomerd, line 300: multiline format

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,322 @@
...
+            self.metricsType = metricsType
+            self.labels = labels
+        }
+    }
+
+    /// Labels for the reported System Metrics.
+    ///
+    /// Backend implementations are encouraged to provide a static extension to this type with
+    /// defaults that suit their specific backend needs.
+    public struct Labels {
```
tomerd, line 144: do we want to provide default labels?

tomerd:
> @tomerd I've addressed your comments from #61 (comment) to the best of my abilities. Would appreciate another round of review/feedback. 😄

@MrLotU the API is taking a great shape. left some comments but I feel we are getting super close to nailing down the API

Sources/CoreMetrics/SystemMetrics.swift
MrLotU: I think in all cases (but I will double check) the `MetricsSystem` Lock is used, so there is no need in this specific context. But a good one to re-check

Sources/CoreMetrics/SystemMetrics.swift
MrLotU: My idea here was to have implementations provide default labels for their backend/style choices. They could then also add default overloads for all the other values.

Personally don't think we should ship labels with this package since it's a very backend specific detail

MrLotU:
Thanks so much @tomerd! I'll go through all your comments tomorrow and apply fixes where needed 😄

Sources/CoreMetrics/Metrics.swift
MrLotU: The reason I did this is so that for the outside world, `MetricsSystem.factory` will still give back the factory they used.

Not doing this would break behaviour like this (used in `SwiftPrometheus`):
```swift
    static func prometheus() throws -> PrometheusClient {
        guard let prom = self.factory as? PrometheusClient else {
            throw PrometheusError.prometheusFactoryNotBootstrapped(bootstrappedWith: "\(self.factory)")
        }
        return prom
    }
```

This also allows us to keep `SystemMetrics` 100% internal other than passing in the `Configuration` at bootstrap time.

Sources/CoreMetrics/SystemMetrics.swift
MrLotU: I think this was leftover from a previous round of work. Will remove it 👍

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,332 @@
...
+            self.timeInterval = config.interval
+            if let dataProvider = config.dataProvider {
+                self.dataProvider = dataProvider
+            } else {
+                #if os(Linux)
+                self.dataProvider = SystemMetrics.linuxSystemMetrics
+                #else
+                self.dataProvider = SystemMetrics.noopSystemMetrics
+                #endif
+            }
```
tomerd, line 54: feels to me like L46 -> L54 can move into the Configuration which can then have its DataProvider member be non-optional, i.e.

```swift
public struct Configuration {
  let interval: DispatchTimeInterval
  let dataProvider: SystemMetrics.DataProvider
  let labels: SystemMetrics.Labels

  public init(pollInterval interval: DispatchTimeInterval = .seconds(2), dataProvider: SystemMetrics.DataProvider? = nil, labels: Labels) {
    self.interval = interval
    if let dataProvider = config.dataProvider {
      self.dataProvider = dataProvider
    } else {
       #if os(Linux)
       self.dataProvider = SystemMetrics.linuxSystemMetrics
       #else
       self.dataProvider = SystemMetrics.noopSystemMetrics
       #endif
    }
    self.labels = labels
  }
}
```

Sources/CoreMetrics/SystemMetrics.swift
tomerd: I think we are assigning self.task in several places which may be unsafe?

MrLotU: There are 2 places where we assign task. One is in the `init`, the other in the `deinit`. You could potentially have a case where the `deinit` is called at the exact same time as the next task execution, but I don't see that happening that quickly.

Wrapping in a lock shouldn't be too much work though, so I'm open to both keeping it this way, as well as locking it :)

tomerd: ah okay, I was under the impression there were more spots but maybe that got removed with cancel :+1:

Sources/CoreMetrics/SystemMetrics.swift
tomerd: @ktoso wdyt?

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,332 @@
...
+        /// Total user and system CPU time spent in seconds.
+        var cpuSeconds: Int
+        /// Maximum number of open file descriptors.
+        var maxFileDescriptors: Int
+        /// Number of open file descriptors.
+        var openFileDescriptors: Int
+    }
+
+    #if os(Linux)
+    internal static func linuxSystemMetrics() -> SystemMetrics.Data? {
```
tomerd, line 202: @drexin @ktoso can you help review the part

tomerd:
@MrLotU thanks for all the hard work getting this done. I am pretty happy with the API shape. I'd let @ktoso help us decide on the default labels question and @drexin and @ktoso make sure the linux imp is correct.

tomerd:

Sources/CoreMetrics/SystemMetrics.swift
drexin: I would suggest using Foundation's `FileManager` instead, which would boil this down to

```swift
guard let statData = FileManager.default.contents(atPath: "/proc/self/stat") else {
  return nil
}

guard let stats = String(data: statData, encoding: .utf8) else {
  return nil
}
```

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,332 @@
...
+        guard
+            let statString = file.readFull()
+            .split(separator: ")")
+            .last
+        else { return nil }
+        let stats = String(statString)
+            .split(separator: " ")
+            .map(String.init)
+        guard
+            let virtualMemoryBytes = Int(stats[safe: 20]),
```
drexin, line 280: I think we should use constants instead of magic numbers here.

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,332 @@
...
+
+        var _rlim = rlimit()
+
+        guard withUnsafeMutablePointer(to: &_rlim, { ptr in
+            getrlimit(__rlimit_resource_t(RLIMIT_NOFILE.rawValue), ptr) == 0
+        }) else { return nil }
+
+        let maxFileDescriptors = Int(_rlim.rlim_max)
+
+        guard let dir = opendir("/proc/self/fd") else { return nil }
```
drexin, line 298: This can also be simplified with `FileManager`:

```swift
let openFileDescriptors = FileManager.default
    .contentsOfDirectory(atPath: "/proc/self/fd")
    .count
```

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,332 @@
...
+        func label(for keyPath: KeyPath<Labels, String>) -> String {
+            return self.prefix + self[keyPath: keyPath]
+        }
+    }
+
+    /// System Metric data.
+    ///
+    /// The current list of metrics exposed is taken from the Prometheus Client Library Guidelines
+    /// https://prometheus.io/docs/instrumenting/writing_clientlibs/#standard-and-runtime-collectors
+    public struct Data {
```
drexin, line 186: This name is not very expressive. Maybe `ProcessStatistics` would be a better name?

Sources/CoreMetrics/SystemMetrics.swift
MrLotU: We used to have an implementation based on `FileManager`, but per @tomerd's request we attempted to keep `CoreMetrics` 100% Foundation free, which is why we went with the C approach.

Sources/CoreMetrics/SystemMetrics.swift
MrLotU: Keep in mind the full name would be `SystemMetrics.Data` since it's a type inside of a different type. Having `SystemMetrics.ProcessStatistics` might become a bit too verbose, if that makes sense 😄 

Agree that `Data` could be improved though 👍

Sources/CoreMetrics/SystemMetrics.swift
```diff
@@ -0,0 +1,337 @@
...
+                Gauge(label: self.labels.label(for: \.maxFileDescriptors)).record(metrics.maxFileDescriptors)
+                Gauge(label: self.labels.label(for: \.openFileDescriptors)).record(metrics.openFileDescriptors)
+            }))
+            
+            timer.schedule(deadline: .now() + self.timeInterval, repeating: self.timeInterval)
+            if #available(OSX 10.12, *) {
+                timer.activate()
+            } else {
+                timer.resume()
+            }
```
MrLotU, line 63: Not sure exactly what the difference between `.activate()` and `.resume()` are here, but they achieve the same result. Not sure if it's acceptable to only use `.resume()` instead of both like this.

CC @drexin

drexin: I think `resume` does a little more in case the source ran before and was suspended, but `resume` works for sources that have never been activated. I'm not sure if there are any downsides to using it in this case.

tomerd:
heya @MrLotU, is this ready for review?

MrLotU:
Yup. Have a couple of outstanding questions, but I think we're in a good place 😄

MrLotU:
@tomerd Any update on this? It's been stale for quite a while 😄

MrLotU:
No problem at all. I can see your points for moving this into a separate project and agree this would be a great way to move forward.

I will hopefully have some time over the coming weekend to get a first draft up as a new module within `swift-metrics` to get an idea of what public API is needed 👍

MrLotU:
Took me a bit longer to find the time, but I've migrated all of SystemMetrics to a new module.

The only change I made to the CoreMetrics module is making the lock public and making `withWriterLock` public. However this might not be required if SystemMetrics would ship with it's own lock implementation.

All other functionality & API stays the same

tomerd:
thanks @MrLotU, how about instead of exploring the lock we expose `withWriterLock` and `withReaderLock`dircetly? this will help not leak the `Lock` type which we prefer to keep internal. Then you can call `self.withWriterLock` instead of `self.lock.withWriterLock` etc?

MrLotU:
That'd be a workable solution 👍 Will implement

ktoso:
Awesome work @MrLotU  -- giving it a final review round then and let's find it a home... :)

ktoso (approved):
This looks good to me!

Discussed with Tom that maybe we can quickly spin out an `swift-metrics-extras` to land this in. Other than _where_ to merge it into this looks good, thanks a lot for the prolonged effort @MrLotU !

ktoso:
Thank you very much @MrLotU!

This is now released as part of the _new_ [**swift metrics extras**](https://github.com/apple/swift-metrics-extras) package! 🎉 🎉 🎉 

Closing the PR here then, Thanks again!

tomerd:
thank you @MrLotU

MrLotU:
Thanks @tomerd and @ktoso for the continued help & guidance. Super happy with how this turned out 😄

PR #64, opened 2020-03-02
Add microseconds to the TimeUnit enum

tomerd:
Motivation:

I have a metrics backend where I need to store timer values in
microseconds. The "preferred display unit" models this exactly, but it
seems to have (inadvertently?) omitted the possibility to express
microseconds. Note that among the timer "report" methods there is
already a microseconds option.

Modifications:

adding an emum case is an API breaking change. so instead of just adding the enum case for microseconds, we use this opportunity to change `TimeUnit` from enum to struct so that we can add more cases in the future without breaking the API.

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -195,11 +195,11 @@ public struct TimeUnit: Equatable {
 
     public static let nanoseconds = TimeUnit(code: .nanoseconds, scaleFromNanoseconds: 1)
     public static let microseconds = TimeUnit(code: .microseconds, scaleFromNanoseconds: 1000)
-    public static let milliseconds = TimeUnit(code: .milliseconds, scaleFromNanoseconds: 1000 * 1000)
-    public static let seconds = TimeUnit(code: .seconds, scaleFromNanoseconds: 1000 * 1000 * 1000)
-    public static let minutes = TimeUnit(code: .minutes, scaleFromNanoseconds: 60 * 1000 * 1000 * 1000)
-    public static let hours = TimeUnit(code: .hours, scaleFromNanoseconds: 60 * 60 * 1000 * 1000 * 1000)
-    public static let days = TimeUnit(code: .days, scaleFromNanoseconds: 60 * 60 * 24 * 1000 * 1000 * 1000)
+    public static let milliseconds = TimeUnit(code: .milliseconds, scaleFromNanoseconds: 1000 * TimeUnit.microseconds.scaleFromNanoseconds)
+    public static let seconds = TimeUnit(code: .seconds, scaleFromNanoseconds: 1000 * TimeUnit.milliseconds.scaleFromNanoseconds)
+    public static let minutes = TimeUnit(code: .minutes, scaleFromNanoseconds: 60 * TimeUnit.seconds.scaleFromNanoseconds)
+    public static let hours = TimeUnit(code: .hours, scaleFromNanoseconds: 60 * TimeUnit.minutes.scaleFromNanoseconds)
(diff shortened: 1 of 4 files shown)


tomerd:
to replace https://github.com/apple/swift-metrics/pull/62

Tests/MetricsTests/TestMetrics.swift
```diff
@@ -166,7 +166,9 @@ internal class TestTimer: TimerHandler, Equatable {
             case .minutes: return (value / 1_000_000_000) * 60
             case .seconds: return value / 1_000_000_000
             case .milliseconds: return value / 1_000_000
+            case .microseconds: return value / 1000
             case .nanoseconds: return value
+            default: preconditionFailure("unknown display unit: \(displayUnit)")
```
weissi: Should we maybe store a conversion to another unit for every new unit? We could for example require that each new unit defines its conversion to nanoseconds so we don’t have to crash if we don’t know it?

MrLotU: Or perhaps as a `default` return `value` without any conversion?

weissi: @MrLotU that would lead to wrong data, no?

MrLotU: Ah. Of course 🤦‍♂ I think in that case your proposed solution is a lot better. Disregard my previous comment.

cburrows: I agree with @weissi , this struct might as well be a Double (or Int64) that expresses how many are in a nanosecond. And if you expose that value then no one has to care about the addition of well named static units ever again.

weissi: Double is a good idea. We could also make “seconds” (as the SI unit) the base and require everything to be able to express itself in seconds if nothing better is available?

tomerd: good idea: https://github.com/apple/swift-metrics/pull/64/commits/1c4103e92d52b05e356e66e3b1dd2b51b3caba0b

Sources/CoreMetrics/Metrics.swift
```diff
@@ -183,19 +183,23 @@ public struct TimeUnit: Equatable {
         case days
     }
 
-    private var code: Code
+    private let code: Code
+    let scaleFromNanoseconds: UInt
```
weissi: This needs to be `UInt64` for 32 bit platforms (or Double)

Sources/CoreMetrics/Metrics.swift
```diff
@@ -172,8 +172,34 @@ public class Gauge: Recorder {
...
+        case microseconds
+        case milliseconds
+        case seconds
+        case minutes
+        case hours
+        case days
+    }
+
+    private let code: Code
+    let scaleFromNanoseconds: UInt64
```
cburrows: Should this be `public`? Otherwise no backing metrics module can use it and would need to know the names.

cburrows: Also good call, `Double` would be weird with `Equatable`.

Sources/CoreMetrics/Metrics.swift
```diff
@@ -172,8 +172,34 @@ public class Gauge: Recorder {
...
+        case nanoseconds
+        case microseconds
+        case milliseconds
+        case seconds
+        case minutes
+        case hours
+        case days
+    }
+
+    private let code: Code
```
ktoso: code reads a bit weird, `encoding`?
```suggestion
    private let encoding: Encoding
```

?

tomerd:
@cburrows @tanner0101 @tachyonics @MrLotU @Mordil @Blejwi @ktoso tagged as 2.0.0

PR #65, opened 2020-03-04
README: recommend 2.x releases

weissi:
Update the README to recommend using 2.x releases whilst still allowing for 1.x releases. Most packages are expected to be compatible with both 1.x and 2.x so I think that's the best we can do.


tomerd:
thanks

PR #82, opened 2020-10-08
+metrics should include their labels when printed to ease debugging

ktoso:
### Motivation:

In some situations one may have a Timer in hand but be unsure if that's the "right one" if developing a library or framework where metrics object creation is a bit smarter than manually ad hoc created.

It is useful to be able to print what labels a metric represents.

### Modifications:

- better description impls

### Result:

- easier to debug swift-metrics in advanced use cases

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -84,6 +84,12 @@ public class Counter {
     }
 }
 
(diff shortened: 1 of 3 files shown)


MrLotU (approved):
Would be helpful for sure in scenarios. 👍

tomerd (approved):

PR #83, opened 2020-10-12
Add Timer.recordInterval method

### LINKED ISSUE #79: Additional sugar: Timer.recordInterval(since: DispatchTime, now: DispatchTime = .now())
Noticed during our work on https://github.com/apple/swift-cluster-membership/pull/70 and other projects this pattern:

 ```swift
extension Timer {
     /// Records time interval between the passed in `since` dispatch time and `now`.
     func recordInterval(since: DispatchTime, now: DispatchTime = .now()) {
         self.recordNanoseconds(now.uptimeNanoseconds - since.uptimeNanoseconds)
     }
 }
```

emerges a lot when a message contains "sent at" or something like that, and then reporting metrics becomes `timer.recordInterval(since: message.sentAt)` leaving the user without any having to dance around math with the time amounts

### PR DESCRIPTION
**Pull request by slashmo**
Add a method to `Timer` allowing to record the difference between a given start `DispatchTime` and a given end `DispatchTime`, defaulting to `.now()`.

### Motivation:

I agree with the reasoning @ktoso provided  in #79 

### Modifications:

- Added `Timer.recordInterval` method
- Added unit test for `Timer.recordInterval`

### Result:

Users recording the difference between a past timestamp and "now" will be able to simplify their code by writing:

```swift
timer.recordInterval(since: start)
```

---

Closes #79

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -368,15 +368,6 @@ public class Timer {
     public func recordSeconds<DataType: BinaryFloatingPoint>(_ duration: DataType) {
         self.recordNanoseconds(Double(duration * 1_000_000_000) < Double(Int64.max) ? Int64(duration * 1_000_000_000) : Int64.max)
     }
-
-    /// Record the time interval (with nanosecond precision) between the passed `since` dispatch time and `end` dispatch time.
-    ///
-    /// - parameters:
-    ///   - since: Start of the interval as `DispatchTime`.
-    ///   - end: End of the interval, defaulting to `.now()`.
-    public func recordInterval(since: DispatchTime, end: DispatchTime = .now()) {
-        self.recordNanoseconds(end.uptimeNanoseconds - since.uptimeNanoseconds)
-    }
 }
 
 extension Timer: CustomStringConvertible {
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Dispatch
+
 // MARK: User API
 
 extension Counter {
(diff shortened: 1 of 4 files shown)


ktoso:

Sources/CoreMetrics/Metrics.swift
```diff
@@ -366,6 +368,15 @@ public class Timer {
     public func recordSeconds<DataType: BinaryFloatingPoint>(_ duration: DataType) {
         self.recordNanoseconds(Double(duration * 1_000_000_000) < Double(Int64.max) ? Int64(duration * 1_000_000_000) : Int64.max)
     }
+
+    /// Record the time interval between the passed `since` dispatch time and `end` dispatch time.
```
ktoso, line 372:
```suggestion
    /// Record the time interval (with nanosecond precision) between the passed `since` dispatch time and `end` dispatch time.
```

slashmo: 👌

ktoso (approved):
This one is popping up _a lot_ since often one stores the start time somewhere and then wants to report when the message with that time "comes back" / "is completed" etc.

In OSS this popped up in our SWIM work, but plenty other times internally as well.

I'm +1 on adding it. I don't really mind the dispatch types usage -- it's not as problematic as foundation is, what do others think?

cc @avolokhov @Lukasa @tomerd

ktoso:
Actually, let's please add the function in the `Metrics` module rather than `CoreMetrics`, there's already precedence for adding convenience there, see: https://github.com/apple/swift-metrics/commit/49431459f961b0a51d1b7946399677c5265b0e96#diff-4aed2fd1274217956ad9c1484f8d53a9

Sources/CoreMetrics/Metrics.swift
```diff
@@ -366,6 +368,15 @@ public class Timer {
...
         self.recordNanoseconds(Double(duration * 1_000_000_000) < Double(Int64.max) ? Int64(duration * 1_000_000_000) : Int64.max)
     }
+
+    /// Record the time interval (with nanosecond precision) between the passed `since` dispatch time and `end` dispatch time.
+    ///
+    /// - parameters:
+    ///   - since: Start of the interval as `DispatchTime`.
+    ///   - end: End of the interval, defaulting to `.now()`.
+    public func recordInterval(since: DispatchTime, end: DispatchTime = .now()) {
+        self.recordNanoseconds(end.uptimeNanoseconds - since.uptimeNanoseconds)
```
ktoso, line 378: can we add this next to "`measure()`" in `Metrics`?

slashmo: On it ✅

tomerd:

tomerd (approved):
thanks

Sources/CoreMetrics/Metrics.swift
```diff
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import Dispatch
```
ktoso, line 15: unnecessary, core metrics imports dispatch without need now

ktoso:
New API so this would be 2.1.0 -- marked the issue as such milsetone.

Sources/CoreMetrics/Metrics.swift
slashmo: @ktoso Thanks for the update & sorry for the inconvenience caused! ✌️

ktoso: Yep no worries 👍

PR #86, opened 2020-11-12
Expose lock functionality

MrLotU:
For full motivation see #61 

### Modifications:

Adds `MetricsSystem.withReaderLock` and `MetricsSystem.withWriterLock`. This allows (third party) packages to use swift-metrics' locking without making the `Lock` types public.

### Result:

This will enable easier extendibility of the swift-metrics systems

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -382,6 +382,22 @@ public enum MetricsSystem {
     fileprivate static var _factory: MetricsFactory = NOOPMetricsHandler.instance
     fileprivate static var initialized = false
 
+    /// Acquire a writer lock for the duration of the given block.
+    ///
+    /// - Parameter body: The block to execute while holding the lock.
+    /// - Returns: The value returned by the block.
+    public static func withWriterLock<T>(_ body: () throws -> T) rethrows -> T {
+        return try self.lock.withWriterLock(body)
+    }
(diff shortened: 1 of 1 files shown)


ktoso:
We only need the writer lock don't we? :) 

Can you remove the withReaderLock? 

then LGTM :)

Sources/CoreMetrics/Metrics.swift
```diff
@@ -382,6 +382,22 @@ public enum MetricsSystem {
...
+        return try self.lock.withWriterLock(body)
+    }
+
+    /// Acquire a reader lock for the duration of the given block.
+    ///
+    /// - Parameter body: The block to execute while holding the lock.
+    /// - Returns: The value returned by the block.
+    public static func withReaderLock<T>(_ body: () throws -> T) rethrows -> T {
+        return try self.lock.withReaderLock(body)
+    }
```
ktoso, line 399: We don't actually need the reader I think -- just the writer for bootstrap right?

MrLotU: For the functionality in #61 we only need the writer, indeed. However I figured for completeness sake and to keep options open I'd add both.

Fine with removing the reader lock as well though 😄

ktoso:
Lovely, let's do it 👍

ktoso (approved):

PR #99, opened 2021-07-29
Introduce FloatingPointCounter

rauhul:
Introduce FloatingPointCounter.

### Motivation:

It is not currently possible to record floating point values via the swift-metrics API even if the metrics backend supports it.

### Modifications:

Adds a `FloatingPointCounter` type to allow users to accumulate non-integral metrics backed by a `FloatingPointCounterHandler`. Introduces a default implementation for creating and destroying `FloatingPointCounterHandler`s for metric backends that do not natively support floating point counters.

On such backends, `FloatingPointCounter` is backed by a `AccumulatingRoundingFloatingPointCounter` which accumulates floating point values internally and record increments to a wrapped `CounterHandler` after crossing integer boundaries.

### Result:

Users can create `FloatingPointCounter`s to record floating point values and get enhanced behavior for backends that support floating point values.

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -12,6 +12,10 @@
 //
 //===----------------------------------------------------------------------===//
 
+// MARK: Testing API
+
+internal var _enableAssertions = true
+
 // MARK: User API
 
 extension Counter {
@@ -87,6 +91,83 @@ public class Counter {
 extension Counter: CustomStringConvertible {
     public var description: String {
         return "Counter(\(self.label), dimensions: \(self.dimensions))"
+    }
+}
+
+extension FloatingPointCounter {
+    /// Create a new `FloatingPointCounter`.
+    ///
+    /// - parameters:
+    ///     - label: The label for the `FloatingPointCounter`.
+    ///     - dimensions: The dimensions for the `FloatingPointCounter`.
+    public convenience init(label: String, dimensions: [(String, String)] = []) {
+        let handler = MetricsSystem.factory.makeFloatingPointCounter(label: label, dimensions: dimensions)
+        self.init(label: label, dimensions: dimensions, handler: handler)
+    }
+
+    /// Signal the underlying metrics library that this FloatingPointCounter will never be updated again.
+    /// In response the library MAY decide to eagerly release any resources held by this `FloatingPointCounter`.
+    @inlinable
+    public func destroy() {
+        MetricsSystem.factory.destroyFloatingPointCounter(self.handler)
+    }
+}
+
+/// A FloatingPointCounter is a cumulative metric that represents a single monotonically increasing FloatingPointCounter whose value can only increase or be reset to zero.
+/// For example, you can use a FloatingPointCounter to represent the number of requests served, tasks completed, or errors.
+/// FloatingPointCounter is not supported by all metrics backends, however a default implementation is provided which accumulates floating point increments and records increments to a standard Counter after crossing integer boundaries.
+///
+/// This is the user-facing FloatingPointCounter API.
+///
+/// Its behavior depends on the `FloatingCounterHandler` implementation.
+public class FloatingPointCounter {
+    @usableFromInline
+    var handler: FloatingPointCounterHandler
+    public let label: String
+    public let dimensions: [(String, String)]
+
+    /// Alternative way to create a new `FloatingPointCounter`, while providing an explicit `FloatingPointCounterHandler`.
+    ///
+    /// - warning: This initializer provides an escape hatch for situations where one must use a custom factory instead of the global one.
+    ///            We do not expect this API to be used in normal circumstances, so if you find yourself using it make sure it's for a good reason.
+    ///
+    /// - SeeAlso: Use `init(label:dimensions:)` to create `FloatingPointCounter` instances using the configured metrics backend.
+    ///
+    /// - parameters:
+    ///     - label: The label for the `FloatingPointCounter`.
+    ///     - dimensions: The dimensions for the `FloatingPointCounter`.
+    ///     - handler: The custom backend.
+    public init(label: String, dimensions: [(String, String)], handler: FloatingPointCounterHandler) {
+        self.label = label
+        self.dimensions = dimensions
+        self.handler = handler
+    }
+
+    /// Increment the FloatingPointCounter.
+    ///
+    /// - parameters:
+    ///     - by: Amount to increment by.
+    @inlinable
+    public func increment<DataType: BinaryFloatingPoint>(by amount: DataType) {
+        self.handler.increment(by: Double(amount))
+    }
+
+    /// Increment the FloatingPointCounter by one.
+    @inlinable
+    public func increment() {
+        self.increment(by: 1.0)
+    }
+
+    /// Reset the FloatingPointCounter back to zero.
+    @inlinable
+    public func reset() {
+        self.handler.reset()
+    }
+}
+
+extension FloatingPointCounter: CustomStringConvertible {
+    public var description: String {
+        return "FloatingPointCounter(\(self.label), dimensions: \(self.dimensions))"
     }
 }
 
@@ -422,6 +503,7 @@ public enum MetricsSystem {
 /// The `MetricsFactory` is the bridge between the `MetricsSystem` and the metrics backend implementation.
 /// `MetricsFactory`'s role is to initialize concrete implementations of the various metric types:
 /// * `Counter` -> `CounterHandler`
+/// * `FloatingPointCounter` -> `FloatingPointCounterHandler`
 /// * `Recorder` -> `RecorderHandler`
 /// * `Timer` -> `TimerHandler`
 ///
@@ -451,6 +533,13 @@ public protocol MetricsFactory {
     ///     - dimensions: The dimensions for the `CounterHandler`.
     func makeCounter(label: String, dimensions: [(String, String)]) -> CounterHandler
 
+    /// Create a backing `FloatingPointCounterHandler`.
+    ///
+    /// - parameters:
+    ///     - label: The label for the `FloatingPointCounterHandler`.
+    ///     - dimensions: The dimensions for the `FloatingPointCounterHandler`.
+    func makeFloatingPointCounter(label: String, dimensions: [(String, String)]) -> FloatingPointCounterHandler
+
     /// Create a backing `RecorderHandler`.
     ///
     /// - parameters:
@@ -473,6 +562,13 @@ public protocol MetricsFactory {
     ///     - handler: The handler to be destroyed.
     func destroyCounter(_ handler: CounterHandler)
 
+    /// Invoked when the corresponding `FloatingPointCounter`'s `destroy()` function is invoked.
+    /// Upon receiving this signal the factory may eagerly release any resources related to this counter.
+    ///
+    /// - parameters:
+    ///     - handler: The handler to be destroyed.
+    func destroyFloatingPointCounter(_ handler: FloatingPointCounterHandler)
+
     /// Invoked when the corresponding `Recorder`'s `destroy()` function is invoked.
     /// Upon receiving this signal the factory may eagerly release any resources related to this recorder.
     ///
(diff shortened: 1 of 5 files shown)


Sources/CoreMetrics/Metrics.swift
```diff
@@ -488,6 +580,62 @@ public protocol MetricsFactory {
...
+    var rawValue: Double = 0
+    var previousValue: Int64 = 0
+    var counterHandler: CounterHandler
+
+    init(counterHandler: CounterHandler) {
+        self.counterHandler = counterHandler
+    }
+
+    func increment(by amount: Double) {
+        // FIXME: does this need a lock?
```
tomerd, line 594: or use an atomic

rauhul: Wouldn't the entire function body need to be atomic, otherwise concurrent increments could mess-up the internal state

Sources/CoreMetrics/Metrics.swift
```diff
@@ -488,6 +580,62 @@ public protocol MetricsFactory {
...
+        rawValue += amount
+        let increment = Int64(rawValue) - previousValue
+        assert(increment >= 0)
+        if increment > 0 {
+            previousValue += increment
+            counterHandler.increment(by: increment)
+        }
+    }
```
tomerd, line 602: we should protect from overflows ^^

tomerd:

rauhul:
Should these types be called `FloatCounter*`?

Sources/CoreMetrics/Metrics.swift
```diff
@@ -488,6 +580,74 @@ public protocol MetricsFactory {
...
+    /// Invoked when the corresponding `FloatingPointCounter`'s `destroy()` function is invoked.
+    /// Upon receiving this signal the factory may eagerly release any resources related to this counter.
+    ///
+    /// `destroyFloatingPointCounter` must be implemented if `makeFloatingPointCounter` is implemented.
+    ///
+    /// - parameters:
+    ///     - handler: The handler to be destroyed.
+    public func destroyFloatingPointCounter(_ handler: FloatingPointCounterHandler) {
+        if let handler = handler as? AccumulatingRoundingFloatingPointCounter {
+            destroyCounter(handler.counterHandler)
```
tomerd, line 646: handler.destroy and then handler.counterHandler can stay private?

Sources/CoreMetrics/Metrics.swift
```diff
@@ -488,6 +580,74 @@ public protocol MetricsFactory {
...
+                self.counterHandler.increment(by: roundedIncrement)
+            }
+        }
+    }
+
+    func reset() {
+        self.lock.withLockVoid {
+            self.rawValue = 0
+            self.roundedValue = 0
+            self.counterHandler.reset()
```
tomerd, line 619: self.counterHandler.reset() outside of the lock?

Sources/CoreMetrics/Metrics.swift
```diff
@@ -488,6 +580,74 @@ public protocol MetricsFactory {
...
 }
 
+/// Wraps a CounterHandler, adding support for incrementing by floating point values by storing an accumulated floating point value and recoding increments to the underlying CounterHandler after crossing integer boundaries.
+private class AccumulatingRoundingFloatingPointCounter: FloatingPointCounterHandler {
+    private static let kAutoStateResetThreshold = Double(Int64.max - (2 << 8))
+
+    private var lock = Lock()
+    private var rawValue: Double = 0
+    private var roundedValue: Int64 = 0
+    private(set) var counterHandler: CounterHandler
```
tomerd, line 590: can this be a let?

Sources/CoreMetrics/Metrics.swift
```diff
@@ -488,6 +580,74 @@ public protocol MetricsFactory {
     func destroyTimer(_ handler: TimerHandler)
 }
 
+/// Wraps a CounterHandler, adding support for incrementing by floating point values by storing an accumulated floating point value and recoding increments to the underlying CounterHandler after crossing integer boundaries.
+private class AccumulatingRoundingFloatingPointCounter: FloatingPointCounterHandler {
+    private static let kAutoStateResetThreshold = Double(Int64.max - (2 << 8))
+
+    private var lock = Lock()
```
tomerd, line 587: can this be a let?

Sources/CoreMetrics/Metrics.swift
rauhul: That could interleave with calls to `increment` though unlikely, right?

Sources/CoreMetrics/Metrics.swift
```diff
@@ -488,6 +580,74 @@ public protocol MetricsFactory {
...
+    func increment(by amount: Double) {
+        assert(amount < AccumulatingRoundingFloatingPointCounter.kAutoStateResetThreshold)
+        self.lock.withLockVoid {
+            let incremented = rawValue + amount
+            if incremented < AccumulatingRoundingFloatingPointCounter.kAutoStateResetThreshold {
+                self.rawValue = incremented
+            } else {
+                self.rawValue = amount
+                self.roundedValue = 0
+            }
```
tomerd, line 605: in other places (e.g. Timer) we compare to Int.max and continue to emit max instead of rolling to 0. not sure what a better approach in this case.

tomerd: also we should add unit test to test normal + overflow behaviors

rauhul: this defers that behavior to the underlying `CounterHandler` and just ensures that the computation in `increment` won't trap or overflow. though looking at this closer, I think the logic is wrong, hmm...

Sources/CoreMetrics/Metrics.swift
tomerd: my is worry about a deadlock if the handler gets stuck for whatever reason. presumably, the handler can handle its own concurrency so we dont need to protect it?

Sources/CoreMetrics/Metrics.swift
rauhul: Are you thinking I should add a destroy method? The destroy instance funcs are only available on the `<Metric>` API not the `<Metrics>Handler` SPI, iiuc.

tomerd: right, since AccumulatingRoundingFloatingPointCounter is an internal type and we check the type here, we can presumably add a non-standard func to it?

rauhul: Well it will need access to the metrics factory so the method would end up looking like:
```swift
    func destroy(factory: MetricsFactory) {
        factory.destroyCounter(counterHandler)
    }

    public func destroyFloatingPointCounter(_ handler: FloatingPointCounterHandler) {
        if let handler = handler as? AccumulatingRoundingFloatingPointCounter {
            handler.destroy(factory: self)
        }
    }
```
I wasn't sure if we should pass the metrics factory to `AccumulatingRoundingFloatingPointCounter`; I'm happy to make this change if you think it makes sense.

tomerd: factory is sort of a s singleton, so maybe `MetricsSystem.factory.destroyCounter(self.counterHandler)` would work? none of them are too elegant, but at least with this the private state of FloatingPointCounterHandler is kept private

rauhul: Pushed this change

Sources/CoreMetrics/Metrics.swift
```diff
@@ -488,6 +580,76 @@ public protocol MetricsFactory {
...
+    /// Create a default backing `FloatingPointCounterHandler` for backends which do not naively support floating point counters.
+    ///
+    /// The created FloatingPointCounterHandler is a wrapper around a backend's CounterHandler which accumulates floating point increments and records increments to an underlying CounterHandler after crossing integer boundaries.
+    ///
+    /// - parameters:
+    ///     - label: The label for the `FloatingPointCounterHandler`.
+    ///     - dimensions: The dimensions for the `FloatingPointCounterHandler`.
+    public func makeFloatingPointCounter(label: String, dimensions: [(String, String)]) -> FloatingPointCounterHandler {
+        return AccumulatingRoundingFloatingPointCounter(
+            counterHandler: self.makeCounter(label: label, dimensions: dimensions))
```
tomerd, line 638: I guess we can also simplify this by moving the counter handler creation into the ctor of AccumulatingRoundingFloatingPointCounter

e.g.

```
 public func makeFloatingPointCounter(label: String, dimensions: [(String, String)]) -> FloatingPointCounterHandler {
  return AccumulatingRoundingFloatingPointCounter(label: label, dimensions: dimensions)
}

private class AccumulatingRoundingFloatingPointCounter: FloatingPointCounterHandler {
     private let counterHandler: CounterHandler

     init(label: String dimension: [String: String]) {
         self.counterHandler = MetricsSystem.factory.makeCounter(label: label, dimension: dimension)
     }
```

Sources/CoreMetrics/Metrics.swift
```diff
@@ -488,6 +580,76 @@ public protocol MetricsFactory {
...
+
+    init(counterHandler: CounterHandler) {
+        self.counterHandler = counterHandler
+    }
+
+    func increment(by amount: Double) {
+        // FIXME: what do?
+        let maxIncrement = Double(Int64.max - (2 << 8))
+        assert((0...maxIncrement).contains(amount))
+        let amount = min(amount, maxIncrement)
```
tomerd, line 597: not exactly sure I understand what the above is validating, its it that amount < Int64.max?

rauhul: I wanted to make sure that doubles passed to this integer rounded counter were representable, the newer version would look like this:
```swift
        let largestRepresentableInt64InFloat64 = Double(2 << 53)
        assert((0...largestRepresentableInt64InFloat64).contains(amount))
        let amount = min(amount, largestRepresentableInt64InFloat64)
```

I'm not sure if this is the right behavior though.

Sources/CoreMetrics/Metrics.swift
rauhul: I don't think that is thread safe:

``` swift
 Time | Thread A                 | Thread B
------+--------------------------+-------------------
    0 | counter.inc(...)         | counter.reset()
    1 | <blocked>                | lock
    2 | <blocked>                | fraction = 0
    3 | <blocked>                | unlock
    4 | lock                     | <not scheduled>
    5 | fraction = x             | <not scheduled>
    6 | handler.inc(...)         | <not scheduled>
    7 | unlock                   | <not scheduled>
    8 | return                   | handler.reset()
    9 |                          | return
```

Depending on the value passed to `inc` and the value in `fraction` you could get incorrect results I think.

tomerd: oh, I think you are worried about correctness, not actual concurrent access, right?

rauhul: Yep, it seems like a correctness issue to not lock around the reset

Sources/CoreMetrics/Metrics.swift
rauhul: This behavior changed, we should probably start a new thread

rauhul:
cc @weissi @ktoso @stephentyrone
thoughts?

weissi:
CC @MrLotU

MrLotU:
Looks like a great feature to have! I just opened https://github.com/MrLotU/SwiftPrometheus/pull/55 as a companion, adding native support to SwiftPrometheus 😄

tomerd:
@rauhul lets add a few tests?

rauhul:
> @rauhul lets add a few tests?

@tomerd just pushed some tests

Sources/CoreMetrics/Metrics.swift
```diff
@@ -12,23 +12,27 @@
...
 //===----------------------------------------------------------------------===//
 
+// MARK: Testing API
+
+internal var _enableAssertions = true
+
 // MARK: User API
 
-extension Counter {
+public extension Counter {
```
tomerd, line 21: these formatting changes are not desired. we prefer to make the methods public not the extension. what version of swift-format did you use?

Sources/CoreMetrics/Metrics.swift
```diff
@@ -12,23 +12,27 @@
 //
 //===----------------------------------------------------------------------===//
 
+// MARK: Testing API
+
+internal var _enableAssertions = true
```
tomerd, line 17: not sure how I feel about this flag. can we get rid of it?

Sources/CoreMetrics/Metrics.swift
rauhul: `0.48.4`

Sources/CoreMetrics/Metrics.swift
rauhul: I don't see a way to test the assertion path without it, do you have a better solution?

Tests/MetricsTests/CoreMetricsTests.swift
```diff
@@ -16,10 +16,19 @@
 import XCTest
 
 class MetricsTests: XCTestCase {
-    func testCounters() throws {
+    let metrics = TestMetrics()
```
tomerd, line 19: prefer little global state as possible

Tests/MetricsTests/MetricsTests.swift
```diff
@@ -17,26 +17,25 @@
 import XCTest
 
 class MetricsExtensionsTests: XCTestCase {
-    func testTimerBlock() throws {
+    let metrics = TestMetrics()
```
tomerd, line 20: ditto re global state

Tests/MetricsTests/TestMetrics.swift
```diff
@@ -154,7 +154,7 @@ internal class TestTimer: TimerHandler, Equatable {
         }
     }
 
-    func retriveValueInPreferredUnit(atIndex i: Int) -> Double {
+    func retrieveValueInPreferredUnit(atIndex i: Int) -> Double {
```
tomerd, line 157: <3

Tests/MetricsTests/MetricsTests.swift
rauhul: I can revert these

Sources/CoreMetrics/Metrics.swift
tomerd: sorry about that :/ can you revert these and use 0.44.6? one swiftPM plugins are there, this kind of issues will go away hopefully.

tomerd: or if you want to keep the local version - run it via the docker setup

Sources/CoreMetrics/Metrics.swift
tomerd: dunno! @weissi @ktoso do you have any? maybe we remove the asserts?

Tests/MetricsTests/MetricsTests.swift
rauhul: fixed in `ae6a0cd`

Sources/CoreMetrics/Metrics.swift
weissi: In NIO we test them in [`NIOCrashTester`](https://github.com/apple/swift-nio/blob/main/Sources/NIOCrashTester/CrashTests%2BEventLoop.swift).

ktoso: I used to have a big test suite for crashes but we removed that a while ago. Nowadays doing so in integration tests only. 
CrashTester sounds like the way to go -- we could "steal" that type I guess.

weissi: yeah, go ahead. Just need to copy the three "driver" files and for every new test suite, you'll need to put an entry into this [dict](https://github.com/apple/swift-nio/blob/main/Sources/NIOCrashTester/CrashTestSuites.swift). It's able to automatically enumerate all `test*` methods, so no need to declare every test. (Here's the [method enumeration code](https://github.com/apple/swift-nio/blob/74b3a8d6a09281ea694ab49a3502012f6c5db99d/Sources/NIOCrashTester/main.swift#L76-L89))

Sources/CoreMetrics/Metrics.swift
rauhul: done, ran with `0.44.6`

tomerd:
this feels ready to me. @ktoso @weissi?

ktoso (commented):
Sorry I was slow to review here, read through and looks solid, thanks a lot @rauhul !

ktoso:
Double checking the API break:

```
20:09:15 Func MetricsFactory.destroyFloatingPointCounter(_:) has been added as a protocol requirement
20:09:15 Func MetricsFactory.makeFloatingPointCounter(label:dimensions:) has been added as a protocol requirement
```

this is fine because we're also adding implementations along with it:

```
extension MetricsFactory {
    public func makeFloatingPointCounter(label: String, dimensions: [(String, String)]) -> FloatingPointCounterHandler {
```

so that error we can ignore.

ktoso (approved):

tomerd:
@weissi any thoughts?

tomerd (approved):

ktoso:
I think we're good here, I'll catch up with Johannes if necessary offline.

weissi (commented):
late LGTM. Definitely not an expert here but looks good

ktoso:
Released as 2.2.0 https://github.com/apple/swift-metrics/releases/tag/2.2.0 FYI @MrLotU 👍

MrLotU:
Native SwiftPrometheus support added in https://github.com/MrLotU/SwiftPrometheus/releases/tag/1.0.0-alpha.14

PR #106, opened 2022-01-18
Import TestMetrics from swift-metrics-extras

ktoso:
After a long chain of "originally developed in" (swift distributed actors -> swift cluster membership -> swift metrics extras) finally moving the test metrics to where they should belong, the metrics repo!

This set of types allows for testing metrics.
The backend can be used in tests to bootstrap and then get metrics of the right label/dimensions and write assertions against them.

TODO:
- port the tests too (from actors, pretty sure we have some)

cc @fabianfett @tomerd


ktoso:
Resolves https://github.com/apple/swift-metrics/issues/107

fabianfett (approved):
Looks good from my side. The only issue I can see is the potential naming conflict, since we now have `MetricsTestUtils` in `swift-metrics` and `swift-metrics-extras`. But since adopters need to declare which package an imported library is comming from this shouldn't be an issue, right?

tomerd:
> The only issue I can see is the potential naming conflict, since we now have MetricsTestUtils in swift-metrics and swift-metrics-extras. But since adopters need to declare which package an imported library is coming from this shouldn't be an issue, right?

I actually think this is a problem.. at least until https://forums.swift.org/t/se-0339-module-aliasing-for-disambiguation/54730 is in place

ktoso:
How about moving back to best name ever -- MetricsTestKit then? :-)

fabianfett:
`MetricsTestKit` it is!

ktoso:

ktoso:
@swift-ci please smoke test

ktoso:

ktoso:
Merging since @fabianfett approved  👍

PR #109, opened 2022-03-24
adopt sendable

tomerd:
motivation: adjust to swift 5.6

changes:
* define sendable shims for protocols and structs that may be used in async context
* refactor Gauge to include rather than inherit Recorder
* adjust tests
* add a test to make sure no warning are emittted


Sources/CoreMetrics/Metrics.swift
```diff
@@ -247,20 +253,85 @@ extension Recorder: CustomStringConvertible {
...
+    @inlinable
+    public func destroy() {
+        self.underlying.destroy()
+    }
+}
+
+/// A gauge is a metric that represents a single numerical value that can arbitrarily go up and down.
+/// Gauges are typically used for measured values like temperatures or current memory usage, but also "counts" that can go up and down, like the number of active threads.
+/// Gauges are modeled as `Recorder` with a sample size of 1 and that does not perform any aggregation.
+public final class Gauge: _SwiftMetricsSendable {
```
Lukasa, line 279: Removing the inheritance makes this a semver major: is that intentional?

tomerd: I was hope we could get away without it since we keep the same public methods signature. any other ideas how to achieve this other than `@unchecked Sendable`?

Lukasa: That's the only one I can think of.

Sources/MetricsTestKit/TestMetrics.swift
```diff
@@ -124,6 +124,11 @@ public final class TestMetrics: MetricsFactory {
     }
 }
 
+#if compiler(>=5.6)
+// TODO: ideally this would not be @unchecked Sendable, but getting warnings even tho we are protecting the state with a lock
+extension TestMetrics: @unchecked Sendable {}
```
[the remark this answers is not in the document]
tomerd: yes - that has been my experience. @DougGregor any advice here?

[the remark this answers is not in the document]
Lukasa: I believe this is working as designed. Getting the Sendable checker to understand locks is, I suspect, a whole ton of work.

Sources/CoreMetrics/Metrics.swift
tomerd: updated this

Sources/CoreMetrics/Metrics.swift
```diff
@@ -12,7 +12,9 @@
 //
```
yim-lee, line 12: nit: update copyright years in these files

yim-lee (approved):

tomerd:
@ktoso @fabianfett thoughts?

Tests/MetricsTests/TestMetrics.swift
```diff
@@ -176,12 +182,21 @@ internal class TestTimer: TimerHandler, Equatable {
...
             self.unlock()
         }
         return body()
     }
 }
+
+// MARK: - Sendable support
+
+#if compiler(>=5.6)
+// ideally we would not be using @unchecked here, but concurrency-safety checks do not recognize locks
```
ktoso, line 198: it was because touching static things right? because just touching class state even warn concurrency should be just ignoring if in a class I think 🤔

Sources/CoreMetrics/Metrics.swift
```diff
@@ -908,3 +924,21 @@ public final class NOOPMetricsHandler: MetricsFactory, CounterHandler, FloatingP
...
+extension Timer: Sendable {}
+// ideally we would not be using @unchecked here, but concurrency-safety checks do not recognize locks
+extension AccumulatingRoundingFloatingPointCounter: @unchecked Sendable {}
+#endif
+
+#if compiler(>=5.6)
+@preconcurrency public protocol _SwiftMetricsSendableProtocol: Sendable {}
+#else
+public protocol _SwiftMetricsSendableProtocol {}
+#endif
```
ktoso, line 944: type dance looks ok

ktoso (commented):
Looks OK AFAICS, worth logging issues about what we could not mark as safe -- AFAIR it was access to statics?

Tests/MetricsTests/TestMetrics.swift
tomerd: no it does warn about touching state of class

tomerd:
lets take https://github.com/apple/swift-metrics/pull/112 first

tomerd:
@yim-lee @weissi @Lukasa @ktoso @fabianfett  this is ready to go. ptal

Sources/CoreMetrics/Metrics.swift
```diff
@@ -928,3 +944,22 @@ public final class NOOPMetricsHandler: MetricsFactory, CounterHandler, FloatingP
...
+#if compiler(>=5.6)
+extension MetricsSystem: Sendable {}
+extension Counter: Sendable {}
+extension FloatingPointCounter: Sendable {}
+// must be @unchecked since Gauge inherits Recorder :(
+extension Recorder: @unchecked Sendable {}
+extension Timer: Sendable {}
+// ideally we would not be using @unchecked here, but concurrency-safety checks do not recognize locks
+extension AccumulatingRoundingFloatingPointCounter: @unchecked Sendable {}
+#endif
```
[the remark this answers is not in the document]
Lukasa: Never mind, god knows what I was thinking when I wrote this.

Lukasa (approved):

PR #110, opened 2022-05-04
Update swift tools version to 5.5

makleso6:
### Motivation:
Xcode warning 
**Conversion to Swift 5 is available**

### Modifications:
package.swift

Update swift tools version
### Result:
No warnings in Xcode when using SPM

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:4.2
+// swift-tools-version:5.5
 //===----------------------------------------------------------------------===//
 //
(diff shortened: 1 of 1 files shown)


Lukasa:
I'm inclined not to merge this directly but to go back to Swift 5.4, to match the NIO project.

tomerd:
thanks for the PR @makleso6 we are deliberately supporting as-old-as-possible versions of Swift in these core API packages. The Xcode "help" message in question sounds like a bug on Xcode side, as the package explicitly declares that it wants to support older versions of Swift.

PR #116, opened 2022-07-20
Make `MetricsFactory` Sendable

fabianfett:
In #109, we likely forgot to require sendability for `MetricsFactory`. Given that `MetricsFactory` is accessed through the global `MetricsSystem.factory`, it clearly must be thread safe.

### Modifications

- Add sendability requirement to `MetricsFactory`

### Result:

`MetricsFactory` is sendable and I can remove `@preconcurrency` from my CoreMetric imports.


ktoso (approved):
That seems right, LGTM thanks @fabianfett

ktoso:
The API checker does not understand Sendable, the change is fine -- merging.

Issue #115, opened 2022-07-20
Move to docc for documentation


ktoso:
The time is now! Let's move to docc docs

PR #118, opened 2022-08-23
+docc prepare for docc publishing

ktoso:
Resolves https://github.com/apple/swift-metrics/issues/115


Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:4.2
+// swift-tools-version:5.7
```
tomerd, line 1: we are still on 5.6 :D

Package@swift-5.6.swift
```diff
@@ -0,0 +1,43 @@
+// swift-tools-version:5.6
```
tomerd, line 1: too soon!

scripts/preview_docc.sh
```diff
@@ -0,0 +1,16 @@
...
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of Swift Metrics API project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+xcrun swift package --disable-sandbox preview-documentation --target "$1"
```
tomerd, line 16: xcrun seems wrong, only available on macOS

Sources/Metrics/Docs.docc/index.md
```diff
@@ -0,0 +1,5 @@
+# ``Metrics``
+
+A Metrics API package for Swift.
+
+Refer to `CoreMetrics` module documentation for the majority of types.
```
yim-lee, line 5: Ah, good idea to have an `index.md` in each module. 👍 

For me I would choose to have the detailed docs in `Metrics` instead because it matches the name of the package, but I'm ok either way.

scripts/preview_docc.sh
yim-lee:
```suggestion
swift package --disable-sandbox preview-documentation --target "$1"
```

yep, can remove `xcrun`.

scripts/soundness.sh
```diff
@@ -86,7 +86,7 @@ for language in swift-or-c bash dtrace; do
   matching_files=( -name '*' )
   case "$language" in
       swift-or-c)
-        exceptions=( -name c_nio_http_parser.c -o -name c_nio_http_parser.h -o -name cpp_magic.h -o -name Package.swift -o -name CNIOSHA1.h -o -name c_nio_sha1.c -o -name ifaddrs-android.c -o -name ifaddrs-android.h)
+        exceptions=( -name c_nio_http_parser.c -o -name c_nio_http_parser.h -o -name cpp_magic.h -o -name Package.swift -o -name "Package@swift-*.swift" -o -name CNIOSHA1.h -o -name c_nio_sha1.c -o -name ifaddrs-android.c -o -name ifaddrs-android.h)
```
yim-lee, line 89:
```suggestion
        exceptions=( -name Package.swift -o -name "Package@swift-*.swift" )
```

yim-lee (approved):
We only need to go up to Swift 5.6 for version-specific manifest for now, otherwise LGTM.

scripts/preview_docc.sh
ktoso: Gotcha thanks

Package@swift-5.6.swift
ktoso: Hah right, thanks! :)

Sources/Metrics/Docs.docc/index.md
ktoso: Hm... i guess better to keep "the docs" in the module which has the types since we can't link to the types otherwise -- the cross module limitation of docc -- 😢

PR #125, opened 2023-05-04
Add Package.resolved to .gitignore

fabianfett:
This is a library target. Having a Package.resolved here doesn't make any sense (which is why we never commited it). To stop git from showing this file, we should add it to `.gitignore`.


ktoso (approved):

ktoso:
Yeah I agree with that 👍

PR #127, opened 2023-05-16
MetricsTestKit: expose all metrics publicly. This allows users to run checks on counters when they might not know the exact label

hamzahrmalik:
Provide functions on TestMetrics to retrieve all counters/timers/recorders

### Motivation:

Users may want to iterate over metrics to perform some complex assertions. For example, they might want to confirm that no metrics have been emitted with a certain dimension

### Modifications:

Added functions for allRecorders, allTimers and allCounters which return arrays of TestRecorder, TestTimer and TestCounter

### DIFF
--- a/Sources/MetricsTestKit/TestMetrics.swift
+++ b/Sources/MetricsTestKit/TestMetrics.swift
@@ -183,7 +183,7 @@ extension TestMetrics {
     }
 
     /// All the counters which have been created and not destroyed
-    var counters: [TestCounter] {
+    public var counters: [TestCounter] {
         let counters = self.lock.withLock {
             self._counters
         }
@@ -220,7 +220,7 @@ extension TestMetrics {
     }
 
     /// All the meters which have been created and not destroyed
-    var meters: [TestMeter] {
+    public var meters: [TestMeter] {
         let meters = self.lock.withLock {
             self._meters
         }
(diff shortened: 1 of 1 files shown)


ktoso:

ktoso (approved):
In general I think this is fine, as a testing facility.

Thanks for the PR!

tomerd (approved):

tomerd (approved):
are there any tests we need to add?

Sources/MetricsTestKit/TestMetrics.swift
```diff
@@ -155,15 +155,21 @@ extension TestMetrics {
...
         }
-        guard let testCounter = maybeCounter as? TestCounter else {
-            throw TestMetricsError.illegalMetricType(metric: maybeCounter, expected: "\(TestCounter.self)")
-        }
         return testCounter
     }
 
+    /// Use this function to run assertions on which counters have been created
+    /// - Returns: All the counters which have been created and not destroyed
+    func allCounters() throws -> [TestCounter] {
```
FranzBusch, line 166: This is not `public`

FranzBusch (changes requested):
Shouldn't we have a `allGauges` and now also an `allMeters` method?

hamzahrmalik:
> This has a merge conflict now due to the introduced meters on main, could you rebase @hamzahrmalik ?
> 
> Otherwise LGTM so let's merge this and the follow up which adds tests to this 👍

I actually think we should merge #128 first
That one is simply a refactor, and will make this one very simple (just adding public modifiers)

ktoso:
Yeah that works, merged #128 👍

PR #128, opened 2023-05-25
Use the MetricsTestKit in the MetricsTests rather than using a copy of the TestMetrics utilities

hamzahrmalik:
Use the MetricsTestKit in the MetricsTests rather than using a copy of the TestMetrics utilities

### Motivation:

Currently we have a copy of TestMetrics within the test target, which is entirely redundant

### Modifications:

Make the test target rely on the existing MetricsTestKit target. Expand the functionality of MetricsTestKit to fill gaps needed for these existing tests

### Result:

We lose the copy, and we exercise the public MetricsTestKit which gives us some confidence that it works


ktoso (approved):
LGTM thanks!

PR #129, opened 2023-05-26
Make 'values' on TestCounter, TestRecorder and TestMeter public

hamzahrmalik:
Make 'values' on TestCounter, TestRecorder and TestMeter public

### Motivation:

`TestTimer` already has a public `values: [Double]`. We should also expose equivalent properties (with equivalent types) on `TestCounter`, `TestRecorder` and `TestMeter`

### Modifications:

Change the existing `values` property to have a type of `[Int64]` or `[Double]` to be equivalent to the existing `TestTimer.values` which is `[Double]`. Then, make these `values` properties public

### Result:

Users can run assertions such as `values.count == 2` to ensure the expected number of values were seen. Otherwise, they are limited to checking the lastValue only


Sources/MetricsTestKit/TestMetrics.swift
```diff
@@ -486,14 +492,12 @@ public final class TestTimer: TestMetric, TimerHandler, Equatable {
...
     }
 
-    func retrieveValueInPreferredUnit(atIndex i: Int) -> Double {
-        return self.lock.withLock {
-            let value = _values[i].1
-            guard let displayUnit = self.displayUnit else {
-                return Double(value)
-            }
-            return Double(value) / Double(displayUnit.scaleFromNanoseconds)
+    public func retrieveValueInPreferredUnit(atIndex i: Int) -> Double {
```
ktoso, line 495: I wonder if just `valueInPreferredUnit` would be fine, while making this public?

ktoso (approved):
Yeah I think this is okey and in-line with the goal of the testkit 👍 

One naming idea, wdyt?

PR #130, opened 2023-05-26
Run swiftformat

hamzahrmalik:


ktoso:
The diff you're seeing is because you formtted using a newer version of swift format than the CI here uses and the CI does not make those changed...

ktoso:
Let's stick to what the version in the repo does, but we should update it soon I think

PR #132, opened 2023-06-08
add increment and decrement public API to Meter

tomerd:
motivation: expose increment and decrement public APIs

changes:
* add increment and decrement public API to Meter
* improve the implementation of AccumulatingMeter and TestMeter
* add tests for new APIs
* refactor/modernize other tests


Sources/CoreMetrics/Metrics.swift
```diff
@@ -787,6 +817,24 @@ internal final class AccumulatingMeter: MeterHandler {
...
+            return
+        }
+        // - cannot increment by negative values
+        guard amount.sign == .plus else {
+            return
+        }
+        // - cannot increment by zero
+        guard !amount.isZero else {
+            return
+        }
```
FranzBusch, line 836: Should we just silently return here or give the user any signal that he used an invalid value?

tomerd: we could make this throwing, but not sure if that make the call sites more cumbersome, note this is the default implementation, so in most cases the actual backend will drive the behavior

FranzBusch: I don't throwing makes sense here since it makes the API very awkward to use. I was more thinking if we should precondition this but that's probably also not a great idea. Overall, I am fine with just dropping them on the ground but wanted to raise the question.

cc @ktoso @Lukasa @fabianfett

Lukasa: We could `assert` on this to ensure that debug builds fail.

tomerd: that isa. good idea

FranzBusch (approved):
LGTM!

PR #133, opened 2023-06-23
Introducing Timer.record(_ duration:), closes #114

### LINKED ISSUE #114: Make use of Swift 5.7 Duration and friends for Timer
We have time types in 5.7, let's use them in timer convenience APIs

natikgadzhi:
This pull request adds a convenience API to `Timer` that uses Swift 5.7's `Duration`. 

### Motivation:

I'm looking for ways to learn Swift, and it's open source ecosystem by doing, and this looked like a beginner-friendly little addition that @ktoso requested (well, a year ago) in #114.
Closes #114.

### Modifications:

- [x] `Timer.record(_ duration: Duration)` that's available on iOS16 and friends.
- [x] A unit test
- [x] Added the test case to Linux tests
- [x] Soundness check passes.
- [x] Record duration in nanoseconds.

### Where to start the review

- My `@available` check looks weird, I thought `Duration` is available in Swift 5.7, but looks like it's platform-specific. Do they look okay?
- Looks like DocC should pick up the function variant automatically, but I'm happy to add example / other documentation as needed.
- `Duration` conversion to seconds / nanoseconds is lossy and looks a bit janky, but seems like that's how folks are doing it for now?

### DIFF
--- a/Sources/Metrics/Metrics.swift
+++ b/Sources/Metrics/Metrics.swift
@@ -92,21 +92,16 @@ extension Timer {
     /// - Throws: `TimerError.durationToIntOverflow` if conversion from `Duration` to `Int64` of Nanoseconds overflowed.
     @available(macOS 13, iOS 16, tvOS 15, watchOS 8, *)
     @inlinable
-    public func record(_ duration: Duration) throws {
+    public func record(_ duration: Duration) {
         // `Duration` doesn't have a nice way to convert it nanoseconds or seconds,
-        // so we'll do the multiplication manually.
-        // Nanoseconds are the smallest unit Timer can track, so we'll record in that.
-        let (seconds, overflow) = duration.components.seconds.multipliedReportingOverflow(by: 1_000_000_000)
-        guard !overflow else {
-            throw TimerError.durationToIntOverflow
-        }
+        // and manual conversion can overflow.
+        let seconds = duration.components.seconds.multipliedReportingOverflow(by: 1_000_000_000)
+        guard !seconds.overflow else { return self.recordNanoseconds(Int64.max) }
 
-        let (nanoseconds, attosecondsOverflow) = seconds.addingReportingOverflow(duration.components.attoseconds / 1_000_000_000)
-        guard !attosecondsOverflow else {
-            throw TimerError.durationToIntOverflow
-        }
+        let nanoseconds = seconds.partialValue.addingReportingOverflow(duration.components.attoseconds / 1_000_000_000)
+        guard !nanoseconds.overflow else { return self.recordNanoseconds(Int64.max) }
 
-        self.recordNanoseconds(nanoseconds)
+        self.recordNanoseconds(nanoseconds.partialValue)
     }
 }
 #endif
--- a/Sources/Metrics/Metrics.swift
+++ b/Sources/Metrics/Metrics.swift
@@ -74,3 +74,16 @@ extension Timer {
         }
     }
 }
+
+extension Timer {
+    /// Convenience for recording a duration based on ``Duration``.
+    ///
+    /// - parameters:
+    ///     - duration: The duration to record.
+    @available(macOS 13, iOS 16, tvOS 15, watchOS 8, *)
+    @inlinable
+    public func record(_ duration: Duration) {
+        let durationSeconds = Double(duration.components.seconds) + Double(duration.components.attoseconds) / 1e18
+        self.recordSeconds(durationSeconds)
+    }
+}
--- a/Sources/Metrics/Metrics.swift
+++ b/Sources/Metrics/Metrics.swift
@@ -78,16 +78,18 @@ extension Timer {
 #if swift(>=5.7)
 extension Timer {
     /// Convenience for recording a duration based on ``Duration``.
+    /// Duration will be recorded in nanoseconds.
     ///
     /// - parameters:
     ///     - duration: The duration to record.
     @available(macOS 13, iOS 16, tvOS 15, watchOS 8, *)
     @inlinable
     public func record(_ duration: Duration) {
-        // `Duration` doesn't have a nice way to convert it back to `Double` of seconds,
-        // so instead we can multiply attoseconds by 1e18 and add the number of seconds to it.
-        let durationSeconds = Double(duration.components.seconds) + Double(duration.components.attoseconds) / 1e18
-        self.recordSeconds(durationSeconds)
+        // `Duration` doesn't have a nice way to convert it nanoseconds or seconds,
+        // so we'll do the multiplication manually.
+        // nanoseconds are the smallest unit Timer can track, so we'll record in that.
+        let durationNanoseconds = duration.components.seconds * 1_000_000_000 + duration.components.attoseconds / 1_000_000_000
+        self.recordNanoseconds(durationNanoseconds)
     }
 }
 #endif
--- a/Sources/Metrics/Metrics.swift
+++ b/Sources/Metrics/Metrics.swift
@@ -76,20 +76,37 @@ extension Timer {
 }
 
 #if swift(>=5.7)
+
+public enum TimerError: Error {
+    case durationToIntOverflow
+}
+
 extension Timer {
     /// Convenience for recording a duration based on ``Duration``.
-    /// Duration will be recorded in nanoseconds.
     ///
-    /// - parameters:
-    ///     - duration: The duration to record.
+    /// `Duration` will be converted to an `Int64` number of nanoseconds, and then recorded with nanosecond precision.
+    ///
+    /// - Parameters:
+    ///     - duration: The `Duration` to record.
+    ///
+    /// - Throws: `TimerError.durationToIntOverflow` if conversion from `Duration` to `Int64` of Nanoseconds overflowed.
     @available(macOS 13, iOS 16, tvOS 15, watchOS 8, *)
     @inlinable
-    public func record(_ duration: Duration) {
+    public func record(_ duration: Duration) throws {
         // `Duration` doesn't have a nice way to convert it nanoseconds or seconds,
         // so we'll do the multiplication manually.
-        // nanoseconds are the smallest unit Timer can track, so we'll record in that.
-        let durationNanoseconds = duration.components.seconds * 1_000_000_000 + duration.components.attoseconds / 1_000_000_000
-        self.recordNanoseconds(durationNanoseconds)
+        // Nanoseconds are the smallest unit Timer can track, so we'll record in that.
+        let (seconds, overflow) = duration.components.seconds.multipliedReportingOverflow(by: 1_000_000_000)
+        guard !overflow else {
+            throw TimerError.durationToIntOverflow
+        }
+
+        let (nanoseconds, attosecondsOverflow) = seconds.addingReportingOverflow(duration.components.attoseconds / 1_000_000_000)
+        guard !attosecondsOverflow else {
+            throw TimerError.durationToIntOverflow
+        }
+
+        self.recordNanoseconds(nanoseconds)
     }
 }
 #endif
--- a/Sources/Metrics/Metrics.swift
+++ b/Sources/Metrics/Metrics.swift
@@ -76,11 +76,6 @@ extension Timer {
 }
 
 #if swift(>=5.7)
-
-public enum TimerError: Error {
-    case durationToIntOverflow
-}
-
 extension Timer {
     /// Convenience for recording a duration based on ``Duration``.
     ///
--- a/Sources/Metrics/Metrics.swift
+++ b/Sources/Metrics/Metrics.swift
@@ -75,6 +75,7 @@ extension Timer {
     }
 }
 
+#if swift(>=5.7)
 extension Timer {
     /// Convenience for recording a duration based on ``Duration``.
     ///
@@ -83,7 +84,10 @@ extension Timer {
     @available(macOS 13, iOS 16, tvOS 15, watchOS 8, *)
     @inlinable
     public func record(_ duration: Duration) {
+        // `Duration` doesn't have a nice way to convert it back to `Double` of seconds,
+        // so instead we can multiply attoseconds by 1e18 and add the number of seconds to it.
         let durationSeconds = Double(duration.components.seconds) + Double(duration.components.attoseconds) / 1e18
         self.recordSeconds(durationSeconds)
     }
 }
+#endif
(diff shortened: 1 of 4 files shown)


Sources/Metrics/Metrics.swift
```diff
@@ -74,3 +74,16 @@ extension Timer {
...
+
+extension Timer {
+    /// Convenience for recording a duration based on ``Duration``.
+    ///
+    /// - parameters:
+    ///     - duration: The duration to record.
+    @available(macOS 13, iOS 16, tvOS 15, watchOS 8, *)
+    @inlinable
+    public func record(_ duration: Duration) {
+        let durationSeconds = Double(duration.components.seconds) + Double(duration.components.attoseconds) / 1e18
```
natikgadzhi, line 86: Should I add an explicit comment on why the hell we're dividing by `1e18` here?

ktoso: Yeah explicit comment will be good 👍

Tests/MetricsTests/MetricsTests.swift
```diff
@@ -95,6 +95,28 @@ class MetricsExtensionsTests: XCTestCase {
...
+    func testTimerDuration() throws {
+        guard #available(iOS 16, macOS 13, tvOS 15, watchOS 8, *) else {
+            return
+        }
+        
+        let metrics = TestMetrics()
+        MetricsSystem.bootstrapInternal(metrics)
+        
+        let name = "timer-\(UUID().uuidString)"
+        let duration = Duration(secondsComponent: 3, attosecondsComponent: 123000000000000000)
```
ktoso, line 108: since Duration does not exist in earlier swift versions this will have to be guarded "around" this code, not just with a `guard #available` since that still makes the code compile but it won't compile on old Swifts :)

ktoso:

Tests/MetricsTests/MetricsTests.swift
natikgadzhi: Thank you for a review!

I wrapped the inside of the test in the `#if` check because this test is added to the array of test cases to run on Linux in an auto-generated file. It's easaier to wrap the insides of the func than to work around not adding that test case conditionally in the script.

natikgadzhi:
I see I broke the formatting, one sec, will fix.

natikgadzhi:
@ktoso, fixed the formatting issues in tests, but `swiftformat` with default config has a few more complaints. Won't add them to this PR to keep this isolated. 

How does that look?

Sources/Metrics/Metrics.swift
```diff
@@ -74,3 +74,20 @@ extension Timer {
...
+    ///
+    /// - parameters:
+    ///     - duration: The duration to record.
+    @available(macOS 13, iOS 16, tvOS 15, watchOS 8, *)
+    @inlinable
+    public func record(_ duration: Duration) {
+        // `Duration` doesn't have a nice way to convert it back to `Double` of seconds,
+        // so instead we can multiply attoseconds by 1e18 and add the number of seconds to it.
+        let durationSeconds = Double(duration.components.seconds) + Double(duration.components.attoseconds) / 1e18
+        self.recordSeconds(durationSeconds)
```
ktoso, line 90: Actually, this is _too_ lossy, we have `recordNanoseconds` that this should call into, as it's the most precision we have. With just seconds this would not be very useful for fast operations

Sources/Metrics/Metrics.swift
```diff
@@ -74,3 +74,20 @@ extension Timer {
         }
     }
 }
+
+#if swift(>=5.7)
+extension Timer {
+    /// Convenience for recording a duration based on ``Duration``.
+    ///
```
ktoso, line 81: Maybe add a note that the duration will be recorded at nanosecond precission

ktoso:
Minor nitpick still, looking good though

natikgadzhi:
@ktoso thank you for the review! Cleaned up, added more comments, recording in nanoseconds now.

tomerd (approved):
nice. thanks @natikgadzhi

Sources/Metrics/Metrics.swift
```diff
@@ -74,3 +74,22 @@ extension Timer {
...
+    /// - parameters:
+    ///     - duration: The duration to record.
+    @available(macOS 13, iOS 16, tvOS 15, watchOS 8, *)
+    @inlinable
+    public func record(_ duration: Duration) {
+        // `Duration` doesn't have a nice way to convert it nanoseconds or seconds,
+        // so we'll do the multiplication manually.
+        // nanoseconds are the smallest unit Timer can track, so we'll record in that.
+        let durationNanoseconds = duration.components.seconds * 1_000_000_000 + duration.components.attoseconds / 1_000_000_000
```
tomerd, line 91: can `duration.components.attoseconds ` ever be zero?

natikgadzhi: - **can duration.components.attoseconds ever be zero?** 
  Unlikely, but they can. That shouldn't be a problem, though? I think it's `Int64` divided by `Int64` that is not zero, so it'll just be zero? 
- **should we be using `multipliedReportingOverflow` and `dividedReportingOverflow` to address overflow risk?**
  TIL! I was suspicious about overflows, thank you for pointing out the direction. I'll take a look and use those, happy to address in this or a new PR.

tomerd: >  I'll take a look and use those, happy to address in this or a new PR.

lets add it here to make sure the code is safe

natikgadzhi: @tomerd, here's an attempt: https://github.com/apple/swift-metrics/pull/133/files#diff-19a0ef33862d9079451ca7ba346b734175c62c5ad95fb30d565a41f9ea32a770R95 

- - `dividedReportingOverflow` should not be applicable in here, since we know we're not dividing by zero. 
- `multipliedReportingOverflow` and `addingReportingOverflow` will return a `Bool` flag on error. Not a very common way of error handling in Swift, but that's cheaper than making a new `Error` and throwing. So, I considered keeping a similar API, and perhaps providing a convenience func that would be throwing, but decided `throws` provides a more familiar and simple API.

What I don't know is whether there's a performance hit we take for marking the function as throwing, in situations where we are not actually throwing errors. I _think_ it's negligible?

Sources/Metrics/Metrics.swift
```diff
@@ -74,3 +74,39 @@ extension Timer {
...
+    ///
+    /// `Duration` will be converted to an `Int64` number of nanoseconds, and then recorded with nanosecond precision.
+    ///
+    /// - Parameters:
+    ///     - duration: The `Duration` to record.
+    ///
+    /// - Throws: `TimerError.durationToIntOverflow` if conversion from `Duration` to `Int64` of Nanoseconds overflowed.
+    @available(macOS 13, iOS 16, tvOS 15, watchOS 8, *)
+    @inlinable
+    public func record(_ duration: Duration) throws {
```
tomerd, line 95: I am not convinced we want these to throw as it will make the call sites awkward to use. I think we should record Int.max instead as we do in `recordNanoseconds`

tomerd: exactly. we already take that approach in `recordXXX`, eg

```

    @inlinable
    public func recordNanoseconds<DataType: BinaryInteger>(_ duration: DataType) {
        self.recordNanoseconds(duration >= Int64.max ? Int64.max : Int64(duration))
    }

    @inlinable
    public func recordSeconds<DataType: BinaryInteger>(_ duration: DataType) {
        guard duration <= Int64.max else { return self.recordNanoseconds(Int64.max) }

        let result = Int64(duration).multipliedReportingOverflow(by: 1_000_000_000)
        if result.overflow {
            self.recordNanoseconds(Int64.max)
        } else {
            self.recordNanoseconds(result.partialValue)
        }
    }
```

natikgadzhi: @tomerd, pushed up a change.

Performing overflow checks one at a time to not perform the second operation if the first already overflowed.

I don't yet know the separation into `CoreMetrics` and plain `Metrics` directories. How do you split them? My gut feeling was that it's about cross-platform support, band `CoreMetrics` is designed to work everywhere where, only relies on standard library, but `Metrics` uses `Foundation` as well? 

Is it important where that Timer extension lives?

natikgadzhi: Aside: I know an engineer at Apple could put that PR and merge it in under an hour, and I appreciate your patience and guidance, @tomerd.

Sources/Metrics/Metrics.swift
```diff
@@ -74,3 +74,34 @@ extension Timer {
         }
     }
 }
+
+#if swift(>=5.7)
+
+public enum TimerError: Error {
```
tomerd, line 80: can be removed now?

Tests/MetricsTests/MetricsTests.swift
```diff
@@ -96,6 +96,36 @@ class MetricsExtensionsTests: XCTestCase {
         XCTAssertEqual(metrics.timers.count, 1, "timer should have been stored")
     }
 
+    func testTimerDuration() throws {
+        // Wrapping only the insides of the test case so that the generated
+        // tests on Linux in MetricsTests+XCTest don't complain that the func does not exist.
+        #if swift(>=5.7)
+        guard #available(iOS 16, macOS 13, tvOS 15, watchOS 8, *) else {
+            return
```
tomerd, line 104: use XCTSkip?

tomerd:
looks good, a couple of nits and we should merge this!

Tests/MetricsTests/MetricsTests.swift
natikgadzhi: TIL.

Tests/MetricsTests/MetricsTests.swift
```diff
@@ -101,7 +101,7 @@ class MetricsExtensionsTests: XCTestCase {
         // tests on Linux in MetricsTests+XCTest don't complain that the func does not exist.
         #if swift(>=5.7)
         guard #available(iOS 16, macOS 13, tvOS 15, watchOS 8, *) else {
-            return
+            throw XCTSkip("Timer.record(_ duration: Duration) is available on Swift 5.7+")
```
tomerd, line 104: do the same for earlier versions of Swift? this one is for the availability. 

you can probably also encapsulated this into a closure and use `XCTSkipIf`, eg (not tested)

XCTSkip({
 #if swift(>=5.7)
  if  #available(iOS 16, macOS 13, tvOS 15, watchOS 8, *) {
    return false
  } else {
    return true
  }
#else
  return true
#endif
}(), ....)

natikgadzhi: Tried that — `#available` doesn't like to be treated as an expression and only works with `guard` and `if`, seemingly. Didn't try for the language check, though. Let me see if that works.

tomerd: adding an `#else` does not work?

natikgadzhi:
```swift
try! XCTSkipIf({
    #if swift(<5.7)
        return true
    #else
        if #available(iOS 16, macOS 13, tvOS 15, watchOS 8, *) {
            return false
        } else {
            return true
        }
    #endif
}(), "Timer.record(_ duration: Duration) is available on Swift 5.7+")
```

That works! But mixing `#if` and regular `if` looks weird to me. Perhaps two separate for `#available` and `#if swift` are okay?

```swift
#if swift(<5.7)
    throw XCTSkip("Timer.record(_ duration: Duration) is not available on this version of Swift")
#endif

guard #available(iOS 16, macOS 13, tvOS 15, watchOS 8, *) else {
    throw XCTSkip("Timer.record(_ duration: Duration) is not available on this version of Swift")
}
```

Looks more readable? The closure with nested ifs and explicit returns seems too verbose. 

Should I stick with two separate checks?

tomerd: I think we will have to `#if swift(>=5.7)` around the entire body of the test as done now since it used types that are not available in previous versions (build will fail) but we can add an `#else` to throw XCTSkip there

tomerd: note tested, but...

```
func testTimerDuration() throws {
        // Wrapping only the insides of the test case so that the generated
        // tests on Linux in MetricsTests+XCTest don't complain that the func does not exist.
        #if swift(>=5.7)
        guard #available(iOS 16, macOS 13, tvOS 15, watchOS 8, *) else {
            throw XCTSkip("Timer.record(_ duration: Duration) is available on Swift 5.7+")
        }

        let metrics = TestMetrics()
        MetricsSystem.bootstrapInternal(metrics)

        let name = "timer-\(UUID().uuidString)"
        let timer = Timer(label: name)

        let duration = Duration(secondsComponent: 3, attosecondsComponent: 123_000_000_000_000_000)
        let nanoseconds = duration.components.seconds * 1_000_000_000 + duration.components.attoseconds / 1_000_000_000
        timer.record(duration)

        // Record a Duration that would overflow,
        // expect Int64.max to be recorded.
        timer.record(Duration(secondsComponent: 10_000_000_000, attosecondsComponent: 123))

        let testTimer = try metrics.expectTimer(timer)
        XCTAssertEqual(testTimer.values.count, 2, "expected number of entries to match")
        XCTAssertEqual(testTimer.values.first, nanoseconds, "expected value to match")
        XCTAssertEqual(testTimer.values[1], Int64.max, "expected to record Int64.max if Durataion overflows")
        XCTAssertEqual(metrics.timers.count, 1, "timer should have been stored")
        #else
        throw XCTSkip("Timer.record(_ duration: Duration) is available on Swift 5.7+")
        #endif
    }

```

natikgadzhi: Alright! Pushed up.

tomerd (approved):

Tests/MetricsTests/MetricsTests.swift
natikgadzhi: Okay, 

```swift
#if swift(>=5.7)
// ... 
#elseif swift(>=5.2)
throw XCTSkip("Timer.record(_ duration: Duration) is only available on Swift >=5.7")
#endif
```


I think that'd be good enough, Swift 5.1 is not that common anymore anyway?

tomerd:
thanks @natikgadzhi

natikgadzhi:
@tomerd, thank you for your time and patience 🙃

ktoso:
Very nice work, late LGTM :)

PR #135, opened 2023-06-23
adds support for swift-concurrency Duration and async closures for Timer measurements

AndriusA:
extends Timer with support for standard lib Duration type and measurements to relative to Instant on Clocks using Duration

### Motivation:

Use of Timer is currently clunky with async code as any durations need to be manually converted to microseconds (et al)

### Modifications:

Extends Timer with:
- method to record duration since `Clock.Instant` on any `Clock` using `Duration`
- methods that use static continuous/suspending clock by default with corresponding instant types
- methods to measure duration of an async closure with provided `Clock` or default `ContinuousClock`

### Result:

Additional Timer methods available for measuring async work

### DIFF
--- a/Sources/Metrics/Metrics.swift
+++ b/Sources/Metrics/Metrics.swift
@@ -24,6 +24,7 @@ extension Timer {
     ///     - dimensions: The dimensions for the Timer.
     ///     - body: Closure to run & record.
     @inlinable
+    @available(*, deprecated, message: "Please use non-static version on an already created Timer")
     public static func measure<T>(label: String, dimensions: [(String, String)] = [], body: @escaping () throws -> T) rethrows -> T {
         let timer = Timer(label: label, dimensions: dimensions)
         let start = DispatchTime.now().uptimeNanoseconds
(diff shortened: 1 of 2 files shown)


AndriusA:
note: need to remove `Timer.record(duration: Duration)` when https://github.com/apple/swift-metrics/pull/133 lands

FranzBusch:
@AndriusA Can you update this PR so we can start to review it?

Sources/Metrics/Metrics.swift
```diff
@@ -74,3 +75,88 @@ extension Timer {
...
+    }
+}
+
+@available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
+internal extension Swift.Duration {
+    /// The duration represented as nanoseconds, clamped to maximum expressible value.
+    var nanosecondsClamped: Int64 {
+        let components = self.components
+
+        let secondsComponentNanos = components.seconds.multipliedReportingOverflow(by: 1_000_000_000)
```
aluco100: maybe this value could be guarded into a static constant since this never mutates and is being used in both `secondsComponentNanos` & `attosCompononentNanos`

Sources/Metrics/Metrics.swift
```diff
@@ -74,3 +75,88 @@ extension Timer {
...
+        let attosCompononentNanos = components.attoseconds / 1_000_000_000
+        let combinedNanos = secondsComponentNanos.partialValue.addingReportingOverflow(attosCompononentNanos)
+
+        guard
+            !secondsComponentNanos.overflow,
+            !combinedNanos.overflow
+        else {
```
aluco100:
```suggestion
            !combinedNanos.overflow else {
```

PR #136, opened 2023-08-02
Add RewritingMetricsHandler

hamzahrmalik:
Add RewritingMetricsFactory

### Motivation:

Sometimes we need rewrite the names of metrics or dimensions into different names. This could be, for example, to make metrics from a new system match what they were in an older system

### Modifications:

Add a new `MetricsRewriter` protocol, which describes how to rewrite metrics
Add a new `RewritingMetricsHandler`, an implementation of MetricsFactory which takes a `MetricsRewriter` and passes all metrics through it

### Notes
* Added a new MetricsWithDimensions type. We might want to make it possible to also use this type in other places (places which currently take a label and dimensions separately)
* The new handler is generic over the underlying factory. This means you cannot initialise it with `any MetricsFactory`. This is because I expect people to have a concrete type at startup when they would create this handler


ktoso:
I think overall this is a great idea, but we probably need to re-consider if we want this in core or maybe would be better to keep in metrics-extras. I know folks would prefer this in a "1.0 package" but that's something we need to discuss separately.

PR #140, opened 2024-04-22
Add `Timer.measure` methods

FranzBusch:
# Motivation

This PR supersedes https://github.com/apple/swift-metrics/pull/135. The goal is to make it easier to measure asynchronous code when using `Metrics`.

# Modification

This PR does:
- Deprecate the current static method for measuring synchronous code
- Add a new instance method to measure synchronous code
- Add a new instance method to measure asynchronous code

### DIFF
--- a/Sources/Metrics/Metrics.swift
+++ b/Sources/Metrics/Metrics.swift
@@ -112,4 +112,44 @@ extension Timer {
 
         self.recordNanoseconds(nanoseconds.partialValue)
     }
+
+    #if compiler(>=6.0)
+    /// Convenience for measuring duration of a closure.
+    ///
+    /// - Parameters:
+    ///    - clock: The clock used for measuring the duration. Defaults to the continuous clock.
+    ///    - body: The closure to record the duration of.
+    @inlinable
+    @available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
+    public func measure<Result, Failure: Error, Clock: _Concurrency.Clock>(
+        clock: Clock = .continuous,
+        body: () throws(Failure) -> Result
+    ) throws(Failure) -> Result where Clock.Duration == Duration {
+        let start = clock.now
+        defer {
+            self.record(duration: start.duration(to: clock.now))
+        }
+        return try body()
+    }
+
+    /// Convenience for measuring duration of a closure.
+    ///
+    /// - Parameters:
+    ///    - clock: The clock used for measuring the duration. Defaults to the continuous clock.
+    ///    - isolation: The isolation of the method. Defaults to the isolation of the caller.
+    ///    - body: The closure to record the duration of.
+    @inlinable
+    @available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
+    public func measure<Result, Failure: Error, Clock: _Concurrency.Clock>(
+        clock: Clock = .continuous,
+        isolation: isolated (any Actor)? = #isolation,
+        body: () async throws(Failure) -> sending Result
+    ) async throws(Failure) -> sending Result where Clock.Duration == Duration {
+        let start = clock.now
+        defer {
+            self.record(duration: start.duration(to: clock.now))
+        }
+        return try await body()
+    }
+    #endif
 }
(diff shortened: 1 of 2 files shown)


FranzBusch:

Lukasa (approved):

Lukasa:

fabianfett (approved):

Sources/Metrics/Metrics.swift
```diff
@@ -98,5 +99,41 @@ extension Timer {
...
+        body: () throws -> Result
+    ) rethrows -> Result where Clock.Duration == Duration {
+        let start = clock.now
+        defer {
+            self.record(start.duration(to: clock.now))
+        }
+        return try body()
+    }
+
+    /// Convenience for measuring duration of a closure with a provided clock.
```
gjcairo, line 121: Nit: the doc for this method and the non-async version don't match. Maybe remove the last bit to match that one, since we're defaulting the `clock` param anyways?
```suggestion
    /// Convenience for measuring duration of a closure.
```

gjcairo (approved):
LGTM, just a nit about docs.

FranzBusch:
@fabianfett @simonjbeaumont @czechboy0 I dusted off this PR again and brought up to the latest Swift standards. If you could take another look that would be great.

Sources/Metrics/Metrics.swift
```diff
@@ -24,6 +24,9 @@ extension Timer {
     ///     - label: The label for the Timer.
     ///     - dimensions: The dimensions for the Timer.
     ///     - body: Closure to run & record.
+    #if compiler(>=6.0)
```
czechboy0, line 27: Do we think it's okay to have a deprecation based on a compiler version? What's the experience for an adopter who supports the same 3 Swift versions as us, do they have to also provide two solutions?

Also, the new API has an OS requirement, so we should make sure we don't put any adopters into a situation where they can't fix this warning without bumping their OS requirement (which today requires a major bump).

Sources/Metrics/Metrics.swift
```diff
@@ -112,4 +115,44 @@ extension Timer {
...
     }
+
+    #if compiler(>=6.0)
+    /// Convenience for measuring duration of a closure.
+    ///
+    /// - Parameters:
+    ///    - clock: The clock used for measuring the duration. Defaults to the continuous clock.
+    ///    - body: The closure to record the duration of.
+    @inlinable
+    @available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
```
czechboy0, line 126: What's the recommendation for adopters who aren't yet on these OS versions as deployment targets?

Should we consider offering a variant if this method without the bumped OS requirements, which use pre-Clock way to measure time, for packages with older OS deployment targets? Or do we just recommend they write their own?

czechboy0:
> To address your feedback I removed the deprecation. This will allow current adopters or adopter with platform requirements to stay with the existing API.

Thanks, Franz. And the [second question](https://github.com/apple/swift-metrics/pull/140#discussion_r1986172068) - are we okay only offering this new method to packages with the raised OS requirements?

FranzBusch:
> Thanks, Franz. And the [second question](https://github.com/apple/swift-metrics/pull/140#discussion_r1986172068) - are we okay only offering this new method to packages with the raised OS requirements?

Yes, I think that's fine. The other method exists and the intention of the new one is to take advantage of Clock.

Tests/MetricsTests/MetricsTests.swift
```diff
@@ -19,6 +19,7 @@ import XCTest
 @testable import Metrics
 
 class MetricsExtensionsTests: XCTestCase {
+    @available(*, deprecated)
```
czechboy0, line 22: I think this can be removed now?

czechboy0:
> Yes, I think that's fine. The other method exists and the intention of the new one is to take advantage of Clock.

Right, it's not a big inconvenience - it might just be confusing that the instance method requires a newer OS (because it uses Clock), but the static method doesn't. We could add a `@_disfavoredOverload` instance method variant that doesn't use Clock? Up to you.

FranzBusch:
> Right, it's not a big inconvenience - it might just be confusing that the instance method requires a newer OS (because it uses Clock), but the static method doesn't. We could add a `@_disfavoredOverload` instance method variant that doesn't use Clock? Up to you.

I get your point but I really don't think we should add those. Clock is the way forward and there is the existing static method that provides an escape hatch.

czechboy0 (approved):

PR #142, opened 2024-05-08
Support compiling against Musl

simonjbeaumont:
### Motivation

When building for Linux, we would like to be able to build this package on top of Musl, as well as the usual Glibc.

### Modifications

Replace imports of Glibc with conditional imports using the following pattern:

```diff
- import Glibc
+ #if canImport(Glibc)
+ import Glibc
+ #elseif canImport(Musl)
+ import Musl
+ #else
+ #error("Unsupported runtime")
+ #endif
```

### Result

Can compile against Musl.


ktoso:

ktoso:
Confused why source compat is panicing hmmm


```
22:31:51 /* Removed Decls */
22:31:51 Func Timer.record(_:) has been removed
```

ktoso:

simonjbeaumont:
Hard to see how this patch would have an API impact; suspect this is a misfire of the checker since I notice that:
1. This project uses the nightly image for the API checker pipeline; and 
2. The failures from the re-triggered run are different.
  a. In the first run (CI job #230) it has `Func Timer.record(_:) has been removed`.
  b. In the second run (CI job #231) it has nothing listed in the removed decls section, but the checker still exited non-zero.
  
  
I suggest we kick it one more time?

simonjbeaumont:
Hang on... what's this script doing? It errors if the `shasum` of the report file isn't an exact match:

```
        if ! shasum "$report" | grep -q cefc4ee5bb7bcdb7cb5a7747efa178dab3c794d5; then
            echo ERROR
            echo >&2 "=============================="
            echo >&2 "ERROR: public API change in $f"
            echo >&2 "=============================="
            cat >&2 "$report"
            errors=$(( errors + 1 ))
        else
            echo OK
        fi
```

Looks like this is a very old version of the API checker script from NIO, which has since been updated to use `swift package diagnose-api-breaking-changes`. The current version of the script in this repo is hand-rolled. Probably is misfiring because the format of the JSON report has changed subtly in spite of no API change.

Will open a PR to update the API checker CI script. If we land that and then rebase this PR, it should be fine.

simonjbeaumont:
Fixing the CI pipeline in #143

FranzBusch (approved):

ktoso:

PR #144, opened 2024-05-14
Change Timer.record(duration: Duration) signature to avoid source breakage

hamzahrmalik:
The previous signature introduced in #133 was
```swift
public func record(_ duration: Duration)
```

There was already a similar function
```swift
public func record(_ duration: DispatchTimeInterval)
```

This can cause ambiguity for the compiler when using `.` shortcuts because of an overlap in case names, e.g.
```swift
timer.record(.nanoseconds(123)) // <-- ambiguous, is this DispatchTimeInterval or Duration? 
```

This is a source breaking change, which we have released in 2.4.2
Anyone using code like `timer.record(.nanoseconds(123))`, who has a minimum platform version high enough, will encounter a compiler error on code which compiled fine in 2.4.1 and below

The fix I'm proposing is to add an argument label to the new method to avoid ambiguity, but I'm open to alternative fixes here

### DIFF
--- a/Sources/Metrics/Metrics.swift
+++ b/Sources/Metrics/Metrics.swift
@@ -82,11 +82,9 @@ extension Timer {
     ///
     /// - Parameters:
     ///     - duration: The `Duration` to record.
-    ///
-    /// - Throws: `TimerError.durationToIntOverflow` if conversion from `Duration` to `Int64` of Nanoseconds overflowed.
     @available(macOS 13, iOS 16, tvOS 15, watchOS 8, *)
     @inlinable
-    public func record(_ duration: Duration) {
+    public func record(duration: Duration) {
         // `Duration` doesn't have a nice way to convert it nanoseconds or seconds,
         // and manual conversion can overflow.
         let seconds = duration.components.seconds.multipliedReportingOverflow(by: 1_000_000_000)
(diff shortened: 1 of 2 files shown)


gjcairo:
Annotating the old `public func record(_ duration: DispatchTimeInterval)` with `@_disfavoredOverload` is another solution, I believe. When in an ambiguous scenario, we'd default to the new `Duration` overload instead. Not sure which is more preferable.

hamzahrmalik:
> Annotating the old `public func record(_ duration: DispatchTimeInterval)` with `@_disfavoredOverload` is another solution, I believe. When in an ambiguous scenario, we'd default to the new `Duration` overload instead. Not sure which is more preferable.

Or maybe you'd want to disfavour the new one, to keep old code doing what it already did? I guess most users don't care if theyre using DispatchTimeInterval.nanoseconds or Duration.nanoseconds but it seems wrong to change it out under them

gjcairo:
Yeah, that's another option - I assumed we may want to favour `Duration`, however I don't have a strong preference for any alternative, just wanted to bring this up. If we aren't particularly interested in moving to `Duration` then we can merge this change.

ktoso:
Thanks for bringing this up! Tbh we may want to prefer the new api… I’ll think about it more in the morning, thank you!

FranzBusch (approved):

FranzBusch:
I am in favor of using the new api and explicit argument labels. This makes it more clear.

ktoso:
Ok yeah that sounds good to me -- let's just sign off 2.4.2 as "broken" and fix the API in this release.

Issue #166, opened 2025-03-18
Compilation failures on 5.10 with new typed throws


0xTim:
The new typed throws are breaking builds on 5.10, even though they're supposed to be wrapped in a compiler check. 

This is possibly related to https://github.com/swift-server/async-http-client/pull/810.

See https://github.com/vapor/vapor/actions/runs/13922803636/job/38961918498?pr=3314 for an example build failure

0xTim:
This is the issue tracking this on Swift https://github.com/swiftlang/swift/issues/79285

PR #168, opened 2025-03-18
Fix 5.10 compile on Ubuntu 24.04

FranzBusch:
Specifically Swift 5.10 on Intel on Ubuntu Noble (24.04) has a crazy bug which leads to compilation failures in a #if compiler(>=6.0) block: swiftlang/swift#79285 .

This workaround fixes the compilation by changing the whitespace.

Fixes https://github.com/apple/swift-metrics/issues/166

### DIFF
--- a/Sources/Metrics/Metrics.swift
+++ b/Sources/Metrics/Metrics.swift
@@ -11,6 +11,9 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 //===----------------------------------------------------------------------===//
+// swift-format-ignore-file
+// Note: Whitespace changes are used to workaround compiler bug
+// https://github.com/swiftlang/swift/issues/79285
 
 @_exported import CoreMetrics
 import Foundation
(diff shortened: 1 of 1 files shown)


Sources/Metrics/Metrics.swift
```diff
@@ -143,8 +146,9 @@ extension Timer {
     public func measure<Result, Failure: Error, Clock: _Concurrency.Clock>(
         clock: Clock = .continuous,
         isolation: isolated (any Actor)? = #isolation,
-        body: () async throws(Failure) -> sending Result
-    ) async throws(Failure) -> sending Result where Clock.Duration == Duration {
+        // DO NOT FIX THE WHITESPACE IN THE NEXT LINE UNTIL 5.10 IS UNSUPPORTED
+        // https://github.com/swiftlang/swift/issues/79285
+        body: () async throws(Failure) -> sending Result ) async throws(Failure) -> sending Result where Clock.Duration == Duration {
```
gwynne, line 151: As currently written, this doesn't seem to be sufficient to bypass the assertion. I suggest:
```suggestion
        body: () async throws(Failure) -> sending Result) async throws(Failure) -> sending Result where Clock.Duration == Duration {
```

czechboy0 (approved):
Looks good, seems this fixed it in AHC 👍

PR #172, opened 2025-04-24
Allows providing a custom MetricsFactory to Counter and friends

czechboy0:
### Motivation:

To allow for things like parallel testing, it'd be useful if we can explicitly provide a MetricsFactory to the existing Counter/... types.

### Modifications:

Added a `factory: MetricsFactory` parameter to all the initializers of Counter/... types, and kept the existing methods that continue to default to `MetricsSystem.factory`.

### Result:

Adopters can use a custom MetricsFactory explicitly passed in at metric creation time. Existing adopters are not affected, unless you opt in, you continue to use the global factory.

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -23,8 +23,10 @@
 ///
 /// Its behavior depends on the `CounterHandler` implementation.
 public final class Counter {
-    /// ``_handler`` is only public to allow access from `MetricsTestKit`. Do not consider it part of the public API.
+    /// `_handler` and `_factory` are only public to allow access from `MetricsTestKit`.
+    /// Do not consider them part of the public API.
     public let _handler: CounterHandler
+    public let _factory: MetricsFactory
     public let label: String
     public let dimensions: [(String, String)]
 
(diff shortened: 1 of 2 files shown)


Sources/CoreMetrics/Metrics.swift
```diff
@@ -23,8 +23,10 @@
 ///
 /// Its behavior depends on the `CounterHandler` implementation.
 public final class Counter {
-    /// ``_handler`` is only public to allow access from `MetricsTestKit`. Do not consider it part of the public API.
+    /// `_handler` and `_factory` are only public to allow access from `MetricsTestKit`.
+    /// Do not consider them part of the public API.
     public let _handler: CounterHandler
+    public let _factory: MetricsFactory
```
FranzBusch, line 29: Can't this be `package` instead then?

czechboy0: Yup, I mirrored the `_handler` here, filed an issue to fix it up across the board separately: https://github.com/apple/swift-metrics/issues/173

Sources/CoreMetrics/Metrics.swift
```diff
@@ -39,10 +42,32 @@ public final class Counter {
...
         self.label = label
         self.dimensions = dimensions
         self._handler = handler
+        self._factory = factory
+    }
+
+    /// Alternative way to create a new `Counter`, while providing an explicit `CounterHandler`.
+    ///
+    /// - warning: This initializer provides an escape hatch for situations where one must use a custom factory instead of the global one.
+    ///            We do not expect this API to be used in normal circumstances, so if you find yourself using it make sure it's for a good reason.
```
FranzBusch, line 56: Is this doc comment correct here (Same for the other types)? This init doesn't take a factory so no way to pass a custom factory. I also don't think we need this and both inits are totally valid to use.

czechboy0: The handler is the "factory" here, I just copied the comment from the existing method.

If we want to clean it up, we should do it separately.

FranzBusch (approved):
LGTM!

PR #187, opened 2025-10-24
Migrate from XCTest to Swift Testing

kukushechkin:
Migrating `swift-metrics` from XCTest to Swift Testing

### Motivation

Swift Testing is a modern alternative to XCTest. Swift Server Ecosystem packages should serve as reference implementations for modern Swift features and best practices.

### Modifications

- Almost 1-to-1 migration of test suites, tests, and assertions to Swift Testing syntax with XCTest import removed
- XCTest references removed from documentation, test example updated
- CI contribution guide updated to be on par with other packages
- A new test suite `GlobalMetricsSystemTests` added, verifying global metrics state

### Result

Successfully executes 55 tests with `swift test --disable-xctest`.

### DIFF
--- a/Sources/MetricsTestKit/Docs.docc/index.md
+++ b/Sources/MetricsTestKit/Docs.docc/index.md
@@ -14,7 +14,7 @@ import MetricsTestKit
 import Testing
 
 struct ExampleTests {
-    @Test func recorderWithcustomMetrics() async throws {
+    @Test func recorderWithCustomMetrics() async throws {
         // Create a local metrics object
         let metrics: TestMetrics = TestMetrics()
 


Sources/MetricsTestKit/Docs.docc/index.md
```diff
@@ -9,29 +9,24 @@ This module offers a ``TestMetrics`` type which you use to bootstrap the metrics
...
 
 ```swift
-import XCTest
 import Metrics
 import MetricsTestKit
+import Testing
 
-final class ExampleTests: XCTestCase {
-    var metrics: TestMetrics! = TestMetrics()
+struct ExampleTests {
```
czechboy0: I wonder if we should show both examples, one with explicitly using a MetricsFactory and passing it around, and a second one with bootstrap (the existing one).

Especially for testing reasons, I think it's good to nudge adopters into explicitly passing around MetricsFactory, especially to components that can be unit-tested.

PR #194, opened 2026-02-03
Mapping metrics factory

### LINKED ISSUE #78: Allow including baggage in reporting metrics?
It should be possible to co-relate metrics to a baggage context.

It effectively means adding some baggage items as labels to a metric.

This likely can be done as extensions in some way?

czechboy0:
A utility for renaming labels and dimensions at the metrics factory level.

### Motivation:

Libraries that emit metrics today have no built-in way to let the application customize metric labels or dimensions at
the factory level.

### Modifications:

Introduce a new public type `MappingMetricsFactory` and a convenience method on `MetricsFactory` to allow centralized renaming of all the created metric objects.

### Result:

Libraries like `swift-system-metrics` can be used with various backends requiring different labels naming convention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ before forwarding it to the upstream factory:
 
 ```swift
 // Add a "service" dimension to all metrics created through this factory
-let factory = myMetricsImplementation.mappingLabelsAndDimensions { label, dimensions in
+let factory = myMetricsImplementation.withLabelAndDimensionsMapping { label, dimensions in
     (label, dimensions + [("service", "checkout")])
 }
 let counter = Counter(label: "request_count", dimensions: [("method", "GET")], factory: factory)
@@ -143,10 +143,10 @@ Transforms can also be chained:
 
 ```swift
 let factory = myMetricsImplementation
-    .mappingLabelsAndDimensions { label, dimensions in
+    .withLabelAndDimensionsMapping { label, dimensions in
         ("myapp.\(label)", dimensions)
     }
-    .mappingLabelsAndDimensions { label, dimensions in
+    .withLabelAndDimensionsMapping { label, dimensions in
         (label, dimensions + [("env", "production")])
     }
 ```
--- a/README.md
+++ b/README.md
@@ -124,6 +124,33 @@ recorder.record(100)
 timer.recordMilliseconds(100)
 ```
 
+### Transforming labels and dimensions
+
+When integrating metrics from multiple libraries or subsystems, you may want to add common dimensions
+(such as service name or environment) or rename labels consistently. `MappingMetricsFactory` wraps an
+existing `MetricsFactory` and applies a transformation to the label and dimensions of every metric
+before forwarding it to the upstream factory:
+
+```swift
+// Add a "service" dimension to all metrics created through this factory
+let factory = myMetricsImplementation.mappingLabelsAndDimensions { label, dimensions in
+    (label, dimensions + [("service", "checkout")])
+}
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -192,6 +192,33 @@ recorder.record(100)
 timer.recordMilliseconds(100)
 ```
 
+### Transforming labels and dimensions
+
+When integrating metrics from multiple libraries or subsystems, you may want to add common dimensions
+(such as service name or environment) or rename labels consistently. `MappingMetricsFactory` wraps an
+existing `MetricsFactory` and applies a transformation to the label and dimensions of every metric
+before forwarding it to the upstream factory:
```
ktoso, line 200:
```suggestion
before forwarding it to the downstream factory:
```

I think this means downstream? 🤔  "next in chain" is downstream I'd say

Sources/CoreMetrics/Docs.docc/index.md
```diff
@@ -171,6 +171,33 @@ Group) to the "sandbox level" of the SSWG's incubation process.
 [discussion](https://forums.swift.org/t/discussion-server-metrics-api/) |
 [feedback](https://forums.swift.org/t/feedback-server-metrics-api/)
 
+### Transforming labels and dimensions
+
+When integrating metrics from multiple libraries or subsystems, you may want to add common dimensions
+(such as service name or environment) or rename labels consistently.
+``MappingMetricsFactory`` wraps an existing ``MetricsFactory`` and applies a transformation to the label
+and dimensions of every metric before forwarding it to the upstream factory:
```
ktoso, line 179: Downstream?

Sources/CoreMetrics/MappingMetricsFactory.swift
```diff
@@ -0,0 +1,180 @@
...
+/// ```
+///
+/// - Note: The transformation only affects what the upstream factory receives. The metric object itself
+///   (for example, `Counter.label`, `Counter.dimensions`) retains the original values passed at creation
+///   time. This means the label you see on the metric handle may differ from the label stored in the
+///   backend. Keeping the original label on the metric object helps when debugging local code. This
+///   discrepancy only requires attention when correlating local metric handles with data in a remote
+///   backend.
+public struct MappingMetricsFactory<Upstream: MetricsFactory>: Sendable {
+    private var upstream: Upstream
```
ktoso, line 38: I see, even in code it's upstream... not sure this is right, if we're wrapping a thing and passing along to it, to me, that sounds like downstream (like in streams libraries), no?

ktoso (approved):
LGTM, just finding the upstream/downstream a bit inverse to what i'd expect

README.md
kukushechkin: I would say "higher up in the call chain" or "inner component" is upstream, like in proxies. Downstream is more like "next step in a pipeline". I.e. wrapping something vs appending another step. Swift Configuration is also using the [same terminology](https://github.com/apple/swift-configuration/blob/main/Sources/Configuration/Providers/Wrappers/KeyMappingProvider.swift#L68).

Sources/CoreMetrics/MappingMetricsFactory.swift
kukushechkin: Answered in another comment https://github.com/apple/swift-metrics/pull/194#discussion_r3078441079

Sources/CoreMetrics/Docs.docc/index.md
kukushechkin: Answered in another comment https://github.com/apple/swift-metrics/pull/194#discussion_r3078441079

PR #221, opened 2026-05-05
Stream-based `MetricsFactory` for debug purposes

### LINKED ISSUE #70: Logging metrics factory
When adding metrics to an application or framework it would be useful to have some indication of the metrics emitted during execution to serve as a basic sanity check ("when I do this does metric X get emitted?"). 

Providing a metrics factory which prints to stdout/stderr whenever a value is recorded would be helpful for this case. Is this something we think we should provide in `swift-metrics`?

### LINKED ISSUE #77: Pattern: Document how one might use multiple metrics backends in one process
It should be possible to use multiple backends in a single process.

We should document how one might want to achieve this.

I think this is the case as I've written tests with concurrent systems against metrics... but will need to double check.

### LINKED ISSUE #189: Text-based metrics factory
Consider creating either a stdout or a swift-log-based text-based metrics factory for quick debugging of metrics emission.

kukushechkin:
Add `StreamMetricsFactory` to `MetricsTestKit`.

###  Motivation

During development and debugging it is sometimes convenient to see metric changes in the process output rather than in a remote system (even if it is running on the same machine).

###  Modifications

- Add a public `StreamMetricsFactory` can emit one log line per metric update to the provided stream.
- "Using multiple backends with debug logging" topic added.

###  Result

Users can bootstrap `StreamMetricsFactory` (or multiplex it with a remote backend) to see metric updates flow through their process output.

Closing #189 #70 and #77.


ktoso (changes requested):
I fear this requires some more consensus... We're tried really hard to keep those core packages dependency free; including "between each other". While it may be weird on server to see metrics without log, I could imagine it happening in some other environments perhaps? (Not sure if any of those libs work in embedded though).

We should discuss this on the forums so the community has a chance to chime in imho (including ESG/SSWG).

Tests/MetricsTests/StreamMetricsFactoryTests.swift
```diff
@@ -0,0 +1,273 @@
...
+import Metrics
+import MetricsTestKit
+import Testing
+
+// MARK: - Test infrastructure
+
+/// Captures every string written to it so the emitted lines can be inspected.
+private final class CapturingStream: TextOutputStream, @unchecked Sendable {
+    private let lock = NSLock()
+    private var buffer: String = ""
```
ktoso, line 26: nitpick, just wondering if not easier to store as [String] instead of splitting each time in `lines`?

ktoso (approved):
Awesome, that resolves the dependency concern :-) LGTM

Tests/MetricsTests/StreamMetricsFactoryTests.swift
kukushechkin: Would it always be a `write` per-line?

Issue #98, opened 2021-07-28
[Question] What is the utility of aggregate?


amadeu01:
I look at the code and tests. However, I couldn't figure out the objective of having `aggregate property.
The only test I found that mention that property was the description one.

https://github.com/apple/swift-metrics/blob/99a068b962176073faec01678d0550db38f6a76d/Tests/MetricsTests/CoreMetricsTests.swift#L395-L410

kukushechkin:
`aggregate` flag is a hint for the backend whether it makes sense to aggregate metric value into buckets. We've added a better explanation throughout the code: https://github.com/apple/swift-metrics/pull/223.

PR #223, opened 2026-05-07
Better explain aggregate flag

kukushechkin:
Expand `aggregate` flag explanation.

Closing #98

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -1386,13 +1386,13 @@ public protocol FloatingPointCounterHandler: AnyObject, _SwiftMetricsSendablePro
 /// This type is an implementation detail and should not be used directly, unless implementing your own metrics backend.
 /// To use the SwiftMetrics API, please refer to the documentation of `Recorder`.
 ///
-/// A handler operates in one of two modes, selected by the `aggregate` argument to
-/// ``MetricsFactory/makeRecorder(label:dimensions:aggregate:)``:
+ /// The `aggregate` argument to ``MetricsFactory/makeRecorder(label:dimensions:aggregate:)``
+  /// selects one of two modes for the handler:
 ///
 /// - **Aggregating** (`aggregate: true`): each call to `record(_:)` contributes a sample to a
 ///   distribution that the backend summarizes. This is the default ``Recorder`` behavior.
-/// - **Non-aggregating** (`aggregate: false`): each call to `record(_:)` replaces the previous
-///   value. This is the behavior ``Gauge`` expects.
+/// - **Non-aggregating** (`aggregate: false`): Each call to `record(_:)` replaces the previous value.
+///   This is the behavior ``Gauge`` expects.
 ///
 /// A single handler instance is bound to one mode — the flag is consumed by the factory
 /// when the handler is created, not on each call.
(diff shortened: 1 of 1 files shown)


FranzBusch (approved):

Sources/CoreMetrics/Metrics.swift
```diff
@@ -1386,6 +1386,17 @@ public protocol FloatingPointCounterHandler: AnyObject, _SwiftMetricsSendablePro
 /// This type is an implementation detail and should not be used directly, unless implementing your own metrics backend.
 /// To use the SwiftMetrics API, please refer to the documentation of `Recorder`.
 ///
+/// A handler operates in one of two modes, selected by the `aggregate` argument to
+/// ``MetricsFactory/makeRecorder(label:dimensions:aggregate:)``:
+///
+/// - **Aggregating** (`aggregate: true`): each call to `record(_:)` contributes a sample to a
+///   distribution that the backend summarizes. This is the default ``Recorder`` behavior.
```
heckj:
```suggestion
/// - **Aggregating** (`aggregate: true`): Each call to `record(_:)` contributes a distribution 
/// that the backend summarizes. This is the default ``Recorder`` behavior.
```

Sources/CoreMetrics/Metrics.swift
```diff
@@ -1386,6 +1386,17 @@ public protocol FloatingPointCounterHandler: AnyObject, _SwiftMetricsSendablePro
...
+/// - **Non-aggregating** (`aggregate: false`): each call to `record(_:)` replaces the previous
```
heckj:
```suggestion
/// - **Non-aggregating** (`aggregate: false`): Each call to `record(_:)` replaces the previous value.
```

Sources/CoreMetrics/Metrics.swift
```diff
@@ -1386,6 +1386,17 @@ public protocol FloatingPointCounterHandler: AnyObject, _SwiftMetricsSendablePro
...
+///   value. This is the behavior ``Gauge`` expects.
```
heckj:
```suggestion
///   This is the behavior ``Gauge`` expects.
```

Sources/CoreMetrics/Metrics.swift
```diff
@@ -1386,6 +1386,17 @@ public protocol FloatingPointCounterHandler: AnyObject, _SwiftMetricsSendablePro
 /// This type is an implementation detail and should not be used directly, unless implementing your own metrics backend.
 /// To use the SwiftMetrics API, please refer to the documentation of `Recorder`.
 ///
+/// A handler operates in one of two modes, selected by the `aggregate` argument to
+/// ``MetricsFactory/makeRecorder(label:dimensions:aggregate:)``:
```
heckj:
```suggestion
 /// The `aggregate` argument to ``MetricsFactory/makeRecorder(label:dimensions:aggregate:)``
  /// selects one of two modes for the handler:
```

(rework to active voice and more direct - the "operates own one of two modes" is a passive voice construction)

Sources/CoreMetrics/Metrics.swift
heckj: tweaking capitalization of Each, and making the two options more parallel in how they're structured

heckj (commented):
Our style guidelines for technical docs generally prefers that we prefix with "A Boolean value that indicates" if we're not speaking directly to what the property represents, but I totally get the implied "Whether" means "Boolean value" and am fine with that if you prefer it for the way it reads.

Sources/CoreMetrics/Metrics.swift
kukushechkin: Looks like there is a duplicate of "/// - **Aggregating** (`aggregate: true`):" in the suggested diff. Could you fix it and I would apply the suggestion?

heckj: Oh crap, sorry, Fixing.

heckj (approved):

Issue #43, opened 2019-08-27
Consider defining type alias for dimension `(String, String)`


yim-lee:
### Expected behavior
It should be clear what each `String` in a dimension tuple represents.

### Actual behavior

Saw this code and had to google search to remind myself that dimensions are just key-value pairs:

```
let dimensions: [(String, String)]
```

Please consider defining type alias:

```
typealias Dimension = (key: String, value: String)

let dimensions: [Dimension]
```

### Steps to reproduce

N/A

### If possible, minimal yet complete reproducer code (or URL to code)

N/A

### SwiftMetrics version/commit hash

1.1.0

### Swift & OS version (output of `swift --version && uname -a`)

N/A

Yasumoto:
Took a quick glance here, and [naming this `Dimension` would collide with the type in `Foundation`](https://github.com/apple/swift-corelibs-foundation/blob/3e5366251756d7ffcf289c0faff865ca7dd23417/Foundation/Unit.swift#L195). Maybe we could call it `MetricDimension` ?

This could be a usability win since we could also add some documentation on the typealias, but it'd be a breaking change. Anyone else have thoughts if this would be helpful?

kukushechkin:
API docs now have explanation what are the tuple values. While unnamed tuple type alias is source-compatible with the raw tuple type, it does not improve the discoverability, as users still need to discover which tuple part is what from the API docs. Unfortunately, we cannot change the API to use a named tuples or a struct without introducing a new major version and there is no plans for a new major version of the package.

PR #224, opened 2026-05-07
Improve discoverability of what are dimension tuples

kukushechkin:
Add explanation to `dimensions` arguments of which tuple value is what.

### Motivation:

We have `([String, String])` all over the `swift-metrics`, but never explain which tuple value is what (even if it looks obvious). We cannot change this to a type alias without breaking the API (even if a `typealias` is source-compatible with the type it represents), as suggested in #43. But we can at least explain which tuple value is what.

### Modifications:

`dimensions` argument comments now have `(name, value)` in the description.

### Result:

Dimensions tuple structure is discoverable without digging through the examples.

### DIFF
--- a/Sources/CoreMetrics/Metrics.swift
+++ b/Sources/CoreMetrics/Metrics.swift
@@ -45,7 +45,7 @@ public final class Counter {
     package let _factory: MetricsFactory
     /// The label for the counter.
     public let label: String
-    /// The dimensions for the counter.
+    /// The dimensions for the counter, as `(name, value)` tuples.
     public let dimensions: [(String, String)]
 
     /// Alternative way to create a new counter with an explicit counter handler.
@@ -54,7 +54,7 @@ public final class Counter {
     ///
     /// - parameters:
     ///   - label: The label for the `Counter`.
-    ///   - dimensions: The dimensions for the `Counter`.
+    ///   - dimensions: The dimensions for the `Counter`, as `(name, value)` tuples.
     ///   - handler: The custom backend.
     ///   - factory: The custom metrics factory.
     public init(label: String, dimensions: [(String, String)], handler: CounterHandler, factory: MetricsFactory) {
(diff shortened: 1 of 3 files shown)


Sources/CoreMetrics/Metrics.swift
```diff
@@ -1489,31 +1489,31 @@ public final class MultiplexMetricsHandler: MetricsFactory {
...
     public func makeMeter(label: String, dimensions: [(String, String)]) -> MeterHandler {
         MuxMeter(factories: self.factories, label: label, dimensions: dimensions)
     }
 
     /// Creates a new recorder handler.
     /// - Parameters:
     ///   - label: The label for the `RecorderHandler`.
-    ///   - dimensions: The dimensions for the `RecorderHandler`.
+    ///   - dimensions: The dimensions for the `RecorderHandler`, as `(name, value)` tuples.
     ///   - aggregate: A Boolean value that indicates whether to aggregate values..
```
heckj:
```suggestion
    ///   - aggregate: A Boolean value that indicates whether to aggregate values.
```

unrelated to your change, but since we're here...

Sources/Metrics/Metrics.swift
```diff
@@ -32,7 +32,7 @@ extension Timer {
     ///
     /// - parameters:
     ///     - label: The label for the Timer.
-    ///     - dimensions: The dimensions for the Timer.
+    ///     - dimensions: The dimensions for the Timer, as `(name, value)` tuples.
```
heckj:
```suggestion
    ///     - dimensions: The dimensions for the `Timer`, as `(name, value)` tuples.
```

the other doc comments used code-voice for the type being referenced here, so tacking that in as a suggestion

Sources/Metrics/Metrics.swift
```diff
@@ -48,7 +48,7 @@ extension Timer {
...
```
heckj:
```suggestion
    ///     - dimensions: The dimensions for the `Timer`, as `(name, value)` tuples.
```

the other doc comments used code-voice for the type being referenced here, so tacking that in as a suggestion

heckj (changes requested):
Generally looks great, a few minor tweaks while we're here

Sources/Metrics/Metrics.swift
kukushechkin: Should all instances of "as `(name, value)` tuples." be "as `(name, value)`." — without "tuples"?

heckj (approved):

PR #238, opened 2026-08-05
Support building with CMake

### LINKED ISSUE #232: Support building with CMake
Support CMake build compatible with [SwiftPM](https://github.com/swiftlang/swift-package-manager/blob/main/CMakeLists.txt#L43) similar to [swift-certificates](https://github.com/apple/swift-certificates/blob/main/CMakeLists.txt), [swift-argument-parser](https://github.com/apple/swift-argument-parser/blob/main/CMakeLists.txt), or other SPM dependencies of SwiftPM.

adityasingh2400:
swift-metrics had no CMake build, so CMake-based consumers such as SwiftPM's own build could not pick it up the way they pick up swift-certificates or swift-argument-parser. This adds a CMake build for the three library targets, `CoreMetrics`, `Metrics` and `MetricsTestKit`, laid out the same way as swift-certificates: a root `CMakeLists.txt`, one per target under `Sources`, and `cmake/modules` carrying `SwiftSupport.cmake` plus the exported config template. The targets are compiled with `-package-name swift-metrics` so the `package` declarations in `CoreMetrics` stay visible to `Metrics` and `MetricsTestKit`, matching what SwiftPM passes.

I also wired up the shared `apple/swift-nio/.github/workflows/cmake_tests.yml` job so the file lists cannot drift from the source tree, and added `/build` to `.gitignore` since that is the directory the shared cmake-build script uses.

Verified locally with CMake 4.4 and Ninja on macOS: a shared-library configure and build succeeds for all three targets, a `BUILD_SHARED_LIBS=NO` configure and build succeeds, and `cmake --install` lays down the archives plus the `.swiftmodule` and `.swiftdoc` files under `lib/swift_static/macosx`. The generated source lists were checked against what `scripts/update-cmake-lists.sh` produces for each target.

Fixes #232

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -29,6 +29,20 @@ jobs:
     name: Cxx interop
     uses: apple/swift-nio/.github/workflows/cxx_interop.yml@main
 
+  cmake-lists:
+    name: Check cmake lists
+    uses: apple/swift-nio/.github/workflows/cmake_tests.yml@main
+    with:
+      image: "swift:6.1-noble"
+      update_cmake_lists_config: >-
+        {
+          "targets": [
+            { "name": "CoreMetrics", "type": "source", "exceptions": [] },
+            { "name": "Metrics", "type": "source", "exceptions": [] },
+            { "name": "MetricsTestKit", "type": "source", "exceptions": [] }
+          ]
+        }
+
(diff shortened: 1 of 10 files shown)


kukushechkin (approved):
@adityasingh2400 Thanks for this!

PR #237, opened 2026-07-21
[SMT-0003] Propose portable metric descriptions

anujraja:
## Summary

Proposes `MetricDescriptor` and descriptor-based factory APIs so portable Swift Metrics
libraries can provide human-readable metric descriptions to backends.

## Motivation

Swift Metrics currently passes only labels and dimensions to backends. This leaves
Prometheus `HELP` text and OpenTelemetry instrument descriptions to backend-specific
workarounds. The proposal keeps that metadata with the metric identity without making
it a dimension or increasing cardinality.

## Design

- Adds an implementation-ready `SMT-0003` proposal for the public API, compatibility
  defaults, backend mapping, migration considerations, and regression-test plan.
- Preserves existing `MetricsFactory` implementations through default descriptor-based
  methods that forward their label and dimensions to today's APIs.
- Explicitly defers implementation until the proposals process marks the design ready.

## Validation

- `git diff --cached --check`
- `swift test` (110 tests passed)

Closes #236

Issue #236, opened 2026-07-20
Add "description" field to metrics


kukushechkin:
Add a "description" field that can be mapped to a [`HELP` line in Prometheus](https://prometheus.io/docs/specs/om/open_metrics_spec_2_0/#help), or an [description field in OTLP](https://opentelemetry.io/docs/concepts/signals/metrics/#metric-instruments).

`swift-prometheus` currently has a separate [registry API with `help:`](https://github.com/swift-server/swift-prometheus/blob/main/Sources/Prometheus/PrometheusCollectorRegistry.swift#L217). `swift-otel` has a [dimensions-based workaround](https://github.com/swift-otel/swift-otel/blob/main/Sources/OTel/OTelCore/Metrics/SwiftMetricsFactory/OTLPMetricsFactory.swift#L293).

It would make sense to introduce new `makeXXX` APIs  in `swift-metrics` to take a hypothetical `stuct MetricDescription` (similar in spirit to [`LogEvent` in `swift-log`](https://github.com/apple/swift-log/blob/main/Sources/Logging/Docs.docc/Proposals/SLG-0005-log-event-LogHandler-API.md)), that is easier to evolve and contain more information. And then update exiting backed implementation so users can have a portable code resulting in `HELP` in Prometheus and `description` in OTel. The change is substantial enough to go through the package [Proposals process](https://github.com/apple/swift-metrics/blob/main/Sources/CoreMetrics/Docs.docc/Proposals/Proposals.md).

PR #239, opened 2026-08-21
Add MetricDescriptor and descriptor-based factory APIs (PoC for #236)

ManjunathAnawal:
### Motivation

Relates to #236. Swift Metrics currently passes only a label and dimensions to backends, so there is no portable way to express a Prometheus/OpenMetrics `HELP` line or an OTLP instrument `description`. Today `swift-prometheus` exposes a separate registry API taking `help:`, and `swift-otel` works around it via dimensions.

### Status: proof of concept, not a merge request

Per the review feedback on #237 — "I strongly recommend to provide a PoC with the proposed solution" — this is the PoC implementation to accompany an SMT proposal, rather than something to land ahead of that process. Opening it as a PR so the diff is reviewable and CI runs against it; happy to convert it to a draft, or close it and keep it as a branch link under the proposal's *Implementation* heading, whichever fits the Proposals workflow better.

### Design

- **`MetricDescriptor`** (`Sendable`, `Equatable`) in `CoreMetrics`: `label`, `dimensions`, and an optional `description`, with an evolvable memberwise init — the "hypothetical struct" from the issue, so future fields (unit, etc.) don't require new factory methods.
- **Five new `MetricsFactory` requirements** — `makeCounter(descriptor:)`, `makeFloatingPointCounter(descriptor:)`, `makeMeter(descriptor:)`, `makeRecorder(descriptor:aggregate:)`, `makeTimer(descriptor:)` — **each with a default implementation** that forwards label and dimensions to the existing `makeXXX(label:dimensions:)` methods. Existing backends compile and behave unchanged; descriptor-aware backends override to receive the description.
- **Descriptor-based convenience initializers** on `Counter`, `FloatingPointCounter`, `Meter`, `Gauge`, `Recorder`, and `Timer` (including the `preferredDisplayUnit` variant). No existing signature was changed or removed.
- **`MultiplexMetricsHandler`** forwards descriptors to every sub-factory; **`MappingMetricsFactory`** transforms label/dimensions while preserving the description; **`TestMetrics`** stores the full descriptor (with `label`/`dimensions` becoming computed properties, so existing reads still work) so tests can assert on descriptions.

### Validation

- `swift test` — 132 tests pass (22 new), including coverage for legacy-only factories, descriptor-aware dispatch, multiplex and mapping propagation, dimension dedupe, and the floating-point-counter fallback.
- `swift package diagnose-api-breaking-changes` — **no breaking changes** in `CoreMetrics`, `Metrics`, or `MetricsTestKit`.
- Builds clean under `-warnings-as-errors -require-explicit-sendable`.
- `swift format lint` clean apart from two findings that are pre-existing on `main`.
- `Sources/CoreMetrics/CMakeLists.txt` updated for the `cmake-lists` check added in #238.

### Open questions for review

- Naming: `MetricDescriptor` here, versus `MetricDescription` as floated in the issue.
- Whether `aggregate` belongs inside the descriptor for `Recorder`, rather than staying a separate parameter.
- Whether the default-implementation fallback should be silent, as here, or surface that a description was dropped.

### DIFF
--- a/Sources/CoreMetrics/MetricDescriptor.swift
+++ b/Sources/CoreMetrics/MetricDescriptor.swift
@@ -1,0 +1,295 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Metrics API open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift Metrics API project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift Metrics API project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+// MARK: - MetricDescriptor
+
+/// A description of a metric's identity and metadata.
+///
+/// A `MetricDescriptor` bundles everything needed to create a metric — its label, dimensions,
+/// and optional human-readable metadata such as a ``description`` — into a single value that is
+/// easy to evolve without breaking the ``MetricsFactory`` API.
+///
+/// Backends can map ``description`` to their native concept of metric help text, for example a
+/// `HELP` line in Prometheus/OpenMetrics exposition, or the `description` field of an
+/// OpenTelemetry instrument.
+///
+/// ```swift
+/// let counter = Counter(
+///     descriptor: MetricDescriptor(
+///         label: "http_requests_total",
+///         dimensions: [("method", "GET")],
+///         description: "Total number of HTTP requests received."
+///     )
+/// )
+/// ```
+public struct MetricDescriptor: Sendable {
+    /// The label (name) of the metric.
+    public var label: String
+    /// The dimensions of the metric, as `(name, value)` tuples.
+    public var dimensions: [(String, String)]
+    /// An optional human-readable description of what the metric measures.
+    ///
+    /// Backends are encouraged to surface this as their native help/description metadata,
+    /// for example a Prometheus `HELP` line or an OpenTelemetry instrument description.
+    /// The description is metadata only: it is not a dimension and does not contribute to
+    /// the metric's identity or cardinality.
+    public var description: String?
+
+    /// Create a new metric descriptor.
+    ///
+    /// - parameters:
+    ///   - label: The label (name) of the metric.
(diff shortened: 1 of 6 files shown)


ManjunathAnawal:
Hi @ktoso @FranzBusch @tomerd .
The "PR label" workflow expired before anyone approved it, so it never ran. Could you approve the workflows and add the appropriate semver label (I think minor, as this adds new API)? I'm happy to rebase or push an update if that helps. Thanks!

kukushechkin:
@ManjunathAnawal thank you for working on this! This would be a great addition to a proposal. I would recommend converting this PR to a draft while you're working on the proposal similar to one in #237.

ManjunathAnawal:
> #237

Thanks  @kukushechkin — I see #237 was closed without merging. Should I pick up SMT-0003 for this proposal, or use the next number? I'll open a fresh proposal PR either way, with #239 (once converted to draft) linked as the implementation.

ManjunathAnawal:
Hi @kukushechkin Proposal is up: #245 — reusing the SMT-0003 slot since #237 was
closed without merging. This PR is now in draft; the proposal links it as the
implementation reference.

Sources/CoreMetrics/MetricDescriptor.swift
```diff
@@ -0,0 +1,295 @@
...
+    ///   - description: An optional human-readable description of what the metric measures.
+    public init(label: String, dimensions: [(String, String)] = [], description: String? = nil) {
+        self.label = label
+        self.dimensions = dimensions
+        self.description = description
+    }
+}
+
+extension MetricDescriptor: Equatable {
+    public static func == (lhs: MetricDescriptor, rhs: MetricDescriptor) -> Bool {
```
kukushechkin: This is more of an implementation review, rather that design, but still worth discussing. The current implementation compares dimensions positionally, which is order-sensitive. That's inconsistent with for example `TestMetrics.FullKey`, where we deliberately made dimension comparison order-insensitive and duplicate-tolerant in #244.

Could you:
1. Make `MetricDescriptor.==` operator sort dimensions before comparing.
2. Extract the sorting/comparison logic into one shared package helper in CoreMetrics so that both `MetricDescriptor` and `TestMetrics` use.
3. Add a test mirroring `countersRepeatedDimensionNames` from #244, but for `MetricDescriptor.==` operator.

PR #244, opened 2026-09-06
Make TestMetrics.FullKey tolerate repeated dimension names

lenamonj:
`TestMetrics` traps when a metric is created with a repeated dimension name.

### Motivation:

`TestMetrics.FullKey` hashes and compares its dimensions through `Dictionary(uniqueKeysWithValues:)`, which traps on a repeated name:

```swift
let metrics = TestMetrics()
Counter(label: "requests", dimensions: [("tag", "a"), ("tag", "b")], factory: metrics)
// Fatal error: Duplicate values for key: 'tag'
```

The metrics API accepts that input and every other factory carries it through.

### Modifications:

`hash(into:)` and `==` read the dimensions sorted by name then value: identity stays order-insensitive, both read the same representation, and repeated names are kept. One test, which crashes without the fix.

### Result:

`TestMetrics` records the instrument instead of killing the test process.


kukushechkin (approved):
Thank you @lenamonj!

